PR #67, opened 2015-05-01
Removed "Divisor" from the countNeighbours effect.

ljw1004:
This "Divisor" is apparently not needed (I tested it without the Divisor, and the code ran fine, and nothing in the subsequent transferTable depends on it having been divided).

Is it really necessary? Or is it just a holdover from an earlier testing phase when you visualized the output of countNeighborsEffect directly and didn't delete it?

### DIFF
--- a/tests/ExampleGallery/Shared/GameOfLife.xaml.cs
+++ b/tests/ExampleGallery/Shared/GameOfLife.xaml.cs
@@ -123,8 +123,7 @@ namespace ExampleGallery
             // Step 1: use a convolve matrix to count how many neighbors are alive. This filter
             // also includes the state of the current cell, but with a lower weighting. The result
             // is an arithmetic encoding where (value / 2) indicates how many neighbors are alive,
-            // and (value % 2) is the state of the cell itself. This is divided by 18 to make it
-            // fit within 0-1 color range.
+            // and (value % 2) is the state of the cell itself.
 
             countNeighborsEffect = new ConvolveMatrixEffect
             {
@@ -135,7 +134,6 @@ namespace ExampleGallery
                     2, 2, 2
                 },
 
-                Divisor = 18,
                 BorderMode = EffectBorderMode.Hard,
             };
 


azurecla:
Hi __@ljw1004__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

ljw1004:
I just tested by rebuilding the sample and running GameOfLife, and it worked fine. I only tested it on Win10 10074 on my SurfacePro - maybe you're trying it on Win8.1? or on a different graphics card? And that could be why you're not observing what I saw.

I don't see anything in liveOrDieEffect that's dependent on its input being in the 0-1 range. Instead, the transfer table is dependent on the input being in the 0-17 range (i.e. pre-divisor). That's why the transfer table is an array with 18 elements.

That said, I know nothing about Direct2D other than what I've learned by looking at your samples. I couldn't understand why this Divisor is present, couldn't see where the 0-1 range is consumed downsteam, observed that GameOfLife works fine on Win10 on my SurfacePro even if I delete it, and so sent this PR to you as a way of having you experts evaluate whether it really is needed. Its fine if you reject this PR because you think it is needed!

clandrew:
I was testing on 8.1. And in fact, you're right- it works fine on 10! I verified this, running the exact same (8.1-style) app on each platform. Same Win2D code. So there is some sort of OS change responsible for the difference here. 

The documentation for [D2D discrete transfer](https://msdn.microsoft.com/en-us/library/windows/desktop/hh706329%28v=vs.85%29.aspx) implies that it's expecting the source values ('C') to fall in the 0..1 range. It may be that D2D/DImage changed the behavior for this effect in 10. I'll backlog something dig into this a bit more.

Because this change won't work for 8.1, a platform we need to support, we can't accept it as-is. Feel free to resubmit if you have one that works on 8.1!

shawnhar:
I suspect the difference is that on 8.1 the intermediate result is written out to a rendertarget, thus clamping values outside 0-1 range, while on Win10 the convolve matrix and discrete transfer are flattened into a single pass (due to new D2D optimizations) which can then preserve the greater range.

We should definitely look into this more to make sure we fully understand the change, and that this change in D2D behavior is intentional.

ljw1004:
@clandrew I agree with your reading of the docs for D2D discrete transfer, but those docs as written are definitely incorrect! It should say:

`V = (V0, V1, .. Vn-1)` where N is the number of elements`

That will make the equation it gives correct. (As things stand, they say that N is the number of elements - 1, which makes the equation incorrect).

They probably also need to clarify the edge condition that `if C=1.0 then pick k=N-1`, which is currently unspecified.

PR #135, opened 2015-08-19
Prevent "Non-chronological" Error for Input Data

18padx08:
Prevents "(Input data cannot be processed in the non-chronological order)", since Input_PointerMoved gets fired with the same timestamp for every Pointer which is on the screen at the same time (e.g multitouch movement).

Not sure what to do if multiple fingers for multiple movements are needed (e.g. move multiple objects)

### DIFF
--- a/samples/ExampleGallery/Shared/CustomFonts.xaml.cs
+++ b/samples/ExampleGallery/Shared/CustomFonts.xaml.cs
@@ -138,8 +138,12 @@ namespace ExampleGallery
             args.Handled = true;
         }
 
+        ulong lastTimestamp = 0;
         private void Input_PointerMoved(object sender, PointerEventArgs args)      
         {
+             //we are not interested in multitouch, so interpret multiple events with same timestamp as one
+            if(args.CurrentPoint.Timestamp <= lastTimestamp) return;
+            lastTimestamp = args.CurrentPoint.Timestamp;
             gestureRecognizer.ProcessMoveEvents(args.GetIntermediatePoints());
             args.Handled = true;
         }


azurecla:
Hi __@18padx08__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

shawnhar:
Thanks for the contribution!

Can you explain more about the situation this change is fixing?  What do I need to do to repro the problem? I tried using multitouch with the original custom font example before this change, but was not able to make anything go wrong...

18padx08:
I checked it on the WinPhone 8.1 build, where i put two, or more fingers on the screen and moved them chaotically around the screen. Then I got "(Input data cannot be processed in the non-chronological order.)" Error, actually pretty fast (two or more moves). I tested it on a Lumia 830 with Windows 10 Technical Preview build 10.0.10512.1000.

As soon as I checked for the timestamp, I could not reproduce the error.

But this error seems to be there for longer time:
https://social.msdn.microsoft.com/Forums/windowsapps/en-US/1c3f0737-b2d7-432e-b4c5-50078454af46/exception-thrown-by-gesturerecognizer?forum=winappswithcsharp

shawnhar:
Thanks.   Damyan was able to repro this on his phone as well, although for some reason I still can't make it happen on my (different) test device.

We are going to follow up with the gesture recognizer team about this issue.  It seems like a platform bug rather than issue in the Win2D code, and we're not entirely sure what the best way to fix it is.

18padx08:
Cool thanks for the fast answer!

shawnhar:
Hello again,

Many thanks for reporting this problem.  Input team is now aware of the issue and looking at fixing it in the OS.

That being so, I'm going to reject this pull request as I don't think workarounds for this type of bug belong in Win2D sample code.  But please don't feel rejected - we very much appreciate being made aware of the issue!

18padx08:
Hi thanks a lot for forwarding this issue!

PR #196, opened 2015-12-05
Added Glyph metrics methods to CanvasFontFace

guillerot:


azurecla:
Hi __@guillerot__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

clandrew:
Thanks for submitting this pull request! 
We've had a goal of making our exposure of DirectWrite types more complete, and adding exposure of IDWriteFontFace::GetDesignGlyphMetrics and IDWriteFontFace::GetGdiCompatibleGlyphMetrics would help us toward that.
I'll take a look.

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -76,12 +76,24 @@ namespace Microsoft.Graphics.Canvas.Text
     [version(VERSION)]
     typedef struct CanvasGlyph
     {
-        int Index;
+        UINT16 Index;
```
clandrew: Making this a signed-int type was a deliberate decision; it fits better with WinRT patterns and precedents, even though it may not exactly match what DirectWrite returns. Besidea that, is there some reason to make it an unsigned short??

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -175,6 +187,24 @@ namespace Microsoft.Graphics.Canvas.Text
...
 
+        HRESULT GetGlyphMetrics(
+          [in] UINT32 inputCount,
+          [in, size_is(inputCount)] UINT16* inputElements,
+          [in] boolean isSideways,
+          [out] UINT32* outputCount,
+          [out, size_is(, *outputCount), retval] CanvasGlyphMetrics** outputElements);
+
+        HRESULT GetGdiCompatibleGlyphMetrics(
+          [in] float emSize,
```
clandrew: Elsewhere, we name this parameter 'fontSize'. Can we make this match?

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -175,6 +187,24 @@ namespace Microsoft.Graphics.Canvas.Text
...
+          [in] float pixelsPerDip,
```
clandrew: Elsewhere, we expose a DPI parameter, and convert it to/from pixels-per-dip. Can we make this match?

winrt/lib/text/CanvasFontFace.h
```diff
@@ -122,6 +122,24 @@ namespace ABI { namespace Microsoft { namespace Graphics { namespace Canvas { na
...
+          uint32_t inputCount,
+          uint16_t* inputElements,
+          boolean isSideways,
+          uint32_t* outputCount,
+          CanvasGlyphMetrics** outputElements) override;
+
+        IFACEMETHOD(GetGdiCompatibleGlyphMetrics)(
+          float emSize,
+          float pixelsPerDip,
+          Numerics::Matrix3x2 transform,
```
clandrew: The Numerics:: qualifier isn't necessary.

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -76,12 +76,24 @@ namespace Microsoft.Graphics.Canvas.Text
     [version(VERSION)]
```
clandrew: When we add new APIs, our procedure is to also add documentation. Would you be able to do this? 
Our CONTRIBUTING.md contains some general info about validating changes, and touches on this.
More specifically, to run the doc generation step:
* Run builddocs.cmd from tools\docs. 
* It not succeed (expected) when there are docs missing. Stubs for the missing elements will be written to obj/docs/MissingDocs.xml.
* Copy the stub elements from MissingDocs.xml and paste them in winrt/docsrc/CanvasFontFace.xml alongside the others.
* Fill in the stubs with some documentation text.
* Run builddocs.cmd again; it should succeed.

winrt/lib/text/CanvasFontFace.h
```diff
@@ -122,6 +122,24 @@ namespace ABI { namespace Microsoft { namespace Graphics { namespace Canvas { na
...
+          boolean useGdiNatural,
+          uint32_t inputCount,
+          UINT16* inputElements,
```
damyanp: We try and use the C++ standard typenames wherever possible - uint16_t in this case.

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -76,12 +76,24 @@ namespace Microsoft.Graphics.Canvas.Text
...
+        UINT16 Index;
         float Advance;
         float AdvanceOffset;
         float AscenderOffset;
     } CanvasGlyph;
 
+    [version(VERSION)]
+    typedef struct CanvasGlyphMetrics
+    {
+      float LeftSideBearing;
```
clandrew: This new result struct contains {LeftSideBearing, RightSideBearing, TopSideBearing, BottomSideBearing}. Could it be reworked to use a glyphbox rect instead? (e.g., a Rect type called Bounds)

We've done a similar thing with how we expose IDWriteTextLayout::GetOverhangMetrics. It's a lot more convenient to work with a 'blackbox' rather than some numbers which are deltas from a blackbox.

There's a comment in the [DirectWrite documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/dd368084(v=vs.85).aspx) with a pointer on how to do this.

winrt/lib/text/CanvasFontFace.abi.idl
guillerot: I thought it was a mistake because DirectWrite uses uint16 for indices.
But now that you say it, I see GetGlyphIndices returning an array of Int.
My bad, I will change it back.

winrt/lib/text/CanvasFontFace.abi.idl
guillerot: Will do tomorrow :smile:

azurecla:
__@guillerot__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

clandrew:
Thanks for making all these updates and signing our CLA! I'll take a look at the latest snapshot.

winrt/test.internal/graphics/CanvasTextRendererUnitTests.cpp
```diff
@@ -219,7 +219,7 @@ namespace canvas
 				for (int i = 0; i < 3; ++i)
 				{
 					int inc = i * 4;
-					Assert::AreEqual(inc + 1, glyphs[i].Index);
+          Assert::AreEqual(inc + 1, glyphs[i].Index);
```
clandrew: To make this delta a little bit cleaner, would it be possible to revert this file (no whitespace change)?

winrt/lib/text/CanvasFontFace.cpp
```diff
@@ -666,6 +666,114 @@ IFACEMETHODIMP CanvasFontFace::GetGlyphIndices(
...
+IFACEMETHODIMP CanvasFontFace::GetGlyphMetrics(
+  uint32_t inputCount,
+  int* inputElements,
+  boolean isSideways,
+  uint32_t* outputCount,
+  Rect** outputElements)
+{
+  return ExceptionBoundary(
+    [&]
+  {
```
clandrew: This is more minor, but it appears that you have your editor set to two spaces. In the code, we have a standard of treating tabs as four spaces. Would you be able to make this new code consistent with the rest?
Admittedly, we have some places in our code that violate this rule. However, it's the convention.

winrt/lib/text/CanvasFontFace.cpp
```diff
@@ -666,6 +666,114 @@ IFACEMETHODIMP CanvasFontFace::GetGlyphIndices(
...
+    ThrowIfFailed(GetRealizedFontFace()->GetDesignGlyphMetrics(glyphIndices.data(), inputCount, glyphMetrics.data(), isSideways));
+
+    ComArray<Rect> output(inputCount);
+
+    DWRITE_FONT_METRICS1 metrics;
+    GetRealizedFontFace()->GetMetrics(&metrics);
+
+    for (uint32_t i = 0; i < inputCount; ++i)
+    {
+      {
```
clandrew: It looks like there's an unneeded block, here. Would it be possible to remove it?

winrt/lib/text/CanvasFontFace.cpp
```diff
@@ -666,6 +666,114 @@ IFACEMETHODIMP CanvasFontFace::GetGlyphIndices(
...
+        auto leftDesignSpace = glyphMetrics[i].leftSideBearing;
+        auto topDesignSpace = glyphMetrics[i].topSideBearing;
```
clandrew: Does the top boundary here not need to use the verticalOriginY field of DWRITE_GLYPH_METRICS?

clandrew:
This is PR is really close to being merged. Thanks for applying all our feedback and signing the CLA! These new APIs are really appreciated.

Before that happens, would it be possible to squash the commit history? In general we have a convention of one bug fix / feature / API change per commit. More importantly- to have everything be in a good state, builds and passes all validation at each commit.

guillerot:
Fixes done & commit history squashed

clandrew:
Thanks. I'll take a look.

clandrew:
I see the commit history looks good, and the other fixes too.

There was one outstanding comment about applying verticalOriginY while producing the glyph box. Did you find that this wasn't necessary? Seemed to me like it was.

Also, I don't have access to the history any more, but did your change have a unit test?

guillerot:
I didn't include verticalOriginY in the production of the glyphbox because with LayoutBounds (of CanvasTextLayoutRegion) and baseline (of LineMetrics), I'm able to fit the blackbox over the glyph drawn. Maybe I'm wrong ?

guillerot:
Unit tests added
Fix GlyphBox.Y to :  lineGap + ascent - verticalOriginY + topSideBearing
Commit history squashed

clandrew:
Great! Taking a look.

winrt/test.internal/graphics/CanvasFontFaceUnitTests.cpp
```diff
@@ -180,6 +188,10 @@ TEST_CLASS(CanvasFontFaceTests)
 
         Assert::AreEqual(RO_E_CLOSED, canvasFontFace->GetGlyphIndices(1, &u, &u, &iPointer));
 
+        Assert::AreEqual(E_INVALIDARG, canvasFontFace->GetGlyphMetrics(1, &i, true, &u, &rectPointer));
```
clandrew: Were the unit tests able to run without error? It looks like this test will not pass- it should check against RO_E_CLOSED.

clandrew: Aside from this, all the unit tests pass for me.

clandrew:
Out of general curiosity, no particular action needed- was there some scenario that motivated you to add these two APIs? Some app that uses a custom text renderer, and needs to compute glyph boundaries, maybe? Or just helping us out with general completeness?

If it's helping us out with general completeness- we discussed this change a bit internally and decided it would be more complete if it returned more than just a rect. We lose a bit of information, for example the original advance widths, if we return the glyph black box only. The advance widths returned through this API aren't exactly equivalent to the glyph width returned by, say, GetGlyphRunBounds. There's something uniquely valuable about the extents returned through this API.

So, the route this should ultimately go is return a struct that contains the glyph black box *plus* all the other fields in the original DWRITE_GLYPH_METRICS. It's a little redundant and the black box will likely be more important than the rest, but in the interest of being super complete, this is the way.

You're free to do this, as part of this change, if you want, or- hopefully you won't be too offended if we adjust the API later on.  Perhaps this is already sufficient for enabling things your app needs to do. Either way, not blocking on it.

winrt/test.internal/graphics/CanvasFontFaceUnitTests.cpp
guillerot: Dumb copy/paste mistaste, will fix it

guillerot:
Some app using DirectWrite (and others DirectX APIs) motivated me to add those 2 APIs.
I'm currently using SharpDX but I'm willing to change for Win2D.
Now, why those 2 ? Just need to get the draw bounds of each glyph in a TextLayout (and Runs).

I will change today the returned value by a struct containing the blackbox and all the original fields, all with the same units (em or px)

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -82,6 +82,19 @@ namespace Microsoft.Graphics.Canvas.Text
...
+    typedef struct CanvasGlyphMetrics
+    {
+        float LeftSideBearing;
+        float AdvanceWidth;
+        float RightSideBearing;
+        float TopSideBearing;
+        float AdvanceHeight;
+        float BottomSideBearing;
+        float VerticalOriginY;
+        Windows.Foundation.Rect BlackBox;
```
clandrew: Could you please rename this to DrawBounds? That way, it's consistent with our CanvasTextLayout.DrawBounds.

winrt/docsrc/CanvasFontFace.xml
```diff
@@ -130,6 +130,12 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license
     <member name="M:Microsoft.Graphics.Canvas.Text.CanvasFontFace.GetGlyphIndices(System.UInt32[])">
       <summary>Returns the nominal mapping of UCS4 Unicode code points to glyph indices as defined by the font 'CMAP' table.</summary>
     </member>
+    <member name="M:Microsoft.Graphics.Canvas.Text.CanvasFontFace.GetGlyphMetrics(System.Int32[],System.Boolean)">
+      <summary>Gets the metrics and bounds of the glyphs that would get drawn in em units.</summary>
+    </member>
+    <member name="M:Microsoft.Graphics.Canvas.Text.CanvasFontFace.GetGdiCompatibleGlyphMetrics(System.Single,System.Single,System.Numerics.Matrix3x2,System.Boolean,System.Int32[],System.Boolean)">
+      <summary>Gets the metrics and bounds of the glyphs that would get drawn, compatible with what GDI would produce, in pixel units.</summary>
+    </member>
```
clandrew: These are in em units too, right?

winrt/docsrc/CanvasFontFace.xml
```diff
@@ -280,6 +286,33 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license
     <member name="F:Microsoft.Graphics.Canvas.Text.CanvasUnicodeRange.Last">
       <summary>The last code point in the Unicode range.</summary>
     </member>
+    <member name="T:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics">
+      <summary>Represents the metrics of an individual glyph. The units depend on how the metrics are obtained.</summary>
```
clandrew: Aren't the units always em units?

winrt/docsrc/CanvasFontFace.xml
```diff
@@ -280,6 +286,33 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license
     <member name="F:Microsoft.Graphics.Canvas.Text.CanvasUnicodeRange.Last">
       <summary>The last code point in the Unicode range.</summary>
     </member>
+    <member name="T:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics">
```
shawnhar: Please add an experimental tag to this new type:

<member name="T:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics" Experimental="true">

This will warn people using the API that is is new, not yet had bake time, so might change in breaking ways depending on feedback etc.

winrt/docsrc/CanvasFontFace.xml
```diff
@@ -280,6 +286,33 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license
...
+    </member>
+    <member name="F:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics.BottomSideBearing">
+      <summary>Specifies the vertical distance from the bottom edge of the black box to the advance height. This is positive when the bottom edge of the black box is within the layout box, or negative when the bottom edge of black box overhangs the layout box.</summary>
+    </member>
+    <member name="F:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics.VerticalOriginY">
+      <summary>Specifies the Y coordinate of a glyph's vertical origin, in the font's design coordinate system. The y coordinate of a glyph's vertical origin is the sum of the glyph's top side bearing and the top (that is, yMax) of the glyph's bounding box.</summary>
+    </member>
+    <member name="F:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics.BlackBox">
+      <summary>Specifies the bounding box that encloses the glyph.</summary>
```
clandrew: Could this say, 'Specifies the bounds of the parts of the glyph are covered if the glyph is drawn.'.

winrt/lib/text/CanvasFontFace.abi.idl
```diff
@@ -82,6 +82,19 @@ namespace Microsoft.Graphics.Canvas.Text
...
+    [version(VERSION)]
+    typedef struct CanvasGlyphMetrics
+    {
+        float LeftSideBearing;
+        float AdvanceWidth;
+        float RightSideBearing;
+        float TopSideBearing;
+        float AdvanceHeight;
+        float BottomSideBearing;
+        float VerticalOriginY;
```
shawnhar: I think this should be called just VerticalOrigin.   The "Y" part is implied by it being Vertical.

winrt/docsrc/CanvasFontFace.xml
```diff
@@ -280,6 +286,33 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license
...
+      <summary>Specifies the Y offset from the vertical origin of the current glyph to the vertical origin of the next glyph when writing vertically. Note that the term "origin" by itself denotes the horizontal origin. The vertical origin is different. Its Y coordinate is specified by verticalOriginY value, and its X coordinate is half the advanceWidth to the right of the horizontal origin.</summary>
+    </member>
+    <member name="F:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics.BottomSideBearing">
+      <summary>Specifies the vertical distance from the bottom edge of the black box to the advance height. This is positive when the bottom edge of the black box is within the layout box, or negative when the bottom edge of black box overhangs the layout box.</summary>
+    </member>
+    <member name="F:Microsoft.Graphics.Canvas.Text.CanvasGlyphMetrics.VerticalOriginY">
+      <summary>Specifies the Y coordinate of a glyph's vertical origin, in the font's design coordinate system. The y coordinate of a glyph's vertical origin is the sum of the glyph's top side bearing and the top (that is, yMax) of the glyph's bounding box.</summary>
```
clandrew: This gets converted from design space into em space, doesn't it?

winrt/docsrc/CanvasFontFace.xml
guillerot: Though not because of the multiplication with the font size

clandrew: True- but it isn't right to call these 'pixel units'- if anything, it would be DIPs.

Now that I'm seeing it, could you please make GetGdiCompatibleGlyphMetrics not scale by fontSize. It uses fontSize to compute its result, but it should have consistent semantics with GetGlyphMetrics, which returns units in em space. It's up to the app to scale by the font size; as it may also apply a DPI transform or some scaling from the passed-in transform.

winrt/docsrc/CanvasFontFace.xml
guillerot: Yes

clandrew:
This is looking great. The only remaining things I see are to add the Experimental="true" attribute on the CanvasGlyphMetrics in the docs, and to squash the commit history.

guillerot:
BlackBox changed to DrawBounds
VerticalOriginY changed to VerticalOrigin
Font size multiplication in GetGdiCompatibleGlyphMetrics removed
Documentation fixes

shawnhar:
This all looks good to me.  Many thanks for the contribution - your help is much appreciated!

clandrew:
Looks good to me, also. Thanks so much!

guillerot:
Thanks for your advices and encouragement

PR #223, opened 2016-01-05
Added overloads of CanvasBitmap methods, which operator on IMemoryBufferReference

eadordzhiev:
This pull request adds the overloads of CanvasBitmap.CreateFromBytes, CanvasBitmap.SetPixelBytes and CanvasBitmap.GetPixelBytes methods, which operate on IMemoryReference, and the corresponding unit tests. I believe these might be useful.

I'm not sure if I need to add an additional parameter for the starting index. Does Windows Runtime expose any API, like multiplane textures, which may require passing the starting index?

By the way, thank you for Win2D! This project saved lots of my time.

### DIFF
--- a/winrt/docsrc/CanvasBitmap.xml
+++ b/winrt/docsrc/CanvasBitmap.xml
@@ -170,6 +170,18 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license 
       <summary>Creates a CanvasBitmap from the bytes of the specified buffer, using the specified pixel width/height, DPI and alpha behavior.</summary>
       <remarks>List of <a href="PixelFormats.htm">supported pixel formats</a>.</remarks>
     </member>
+	<member name="M:Microsoft.Graphics.Canvas.CanvasBitmap.CreateFromBytes(Microsoft.Graphics.Canvas.ICanvasResourceCreator,Windows.Foundation.IMemoryBufferReference,System.Int32,System.Int32,Windows.Graphics.DirectX.DirectXPixelFormat)">
+      <summary>Creates a CanvasBitmap from the bytes of the specified buffer, using the specified pixel width/height, premultiplied alpha and default (96) DPI.</summary>
+      <remarks>List of <a href="PixelFormats.htm">supported pixel formats</a>.</remarks>
+    </member>
+    <member name="M:Microsoft.Graphics.Canvas.CanvasBitmap.CreateFromBytes(Microsoft.Graphics.Canvas.ICanvasResourceCreator,Windows.Foundation.IMemoryBufferReference,System.Int32,System.Int32,Windows.Graphics.DirectX.DirectXPixelFormat,System.Single)">
+      <summary>Creates a CanvasBitmap from the bytes of the specified buffer, using the specified pixel width/height, specified DPI, and premultiplied alpha.</summary>
+      <remarks>List of <a href="PixelFormats.htm">supported pixel formats</a>.</remarks>
+    </member>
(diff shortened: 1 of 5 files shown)


azurecla:
Hi __@eadordzhiev__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@eadordzhiev__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

shawnhar:
Hi, and thanks for the contribution!

I'm trying to understand what does IMemoryReference enable that can't be done using IBuffer?  As far as I can see the main difference between the two is that IBuffer is potentially marshallable across process boundaries, while IMemoryReference is not, but that should not be relevant for typical Win2D use cases that are all within a single process.

What's the scenario that motivated you to want these overloads?

eadordzhiev:
Some Windows Runtime APIs like SoftwareBitmap or PerceptionFrame operate on IMemoryReference instead of IBuffer. As for SoftwareBitmap, there already exists the CanvasBitmap.CreateFromSoftwareBitmap, but copying a rectangular region from\to a SoftwareBitmap as fas as I know is not possible without additional allocations.

In my project I had to use MemoryBuffer because the standard implementation of IBuffer, which is `Windows.Storage.Steams.Buffer`, doesn't implement IDisposable, but MemoryReference does.

I had needed deterministic memory management, but as long as I've used C#, I couldn't just nullify it like I would (or wouldn't) normally do in C++/CX, and get it instantly freed. I've done some research and it appears that the working set grows up to ~160MB until it gets GCed vs ~70MB max when you use MemoryBuffer and explicitly free it with Dispose().

shawnhar:
Ok, I'm back from doing some MSDN reading to educate myself about IMemoryBuffer :-)

The SoftwareBitmap usage is interesting in that it has a LockBuffer API which returns an IMemoryBufferReference, but the other buffer related SoftwareBitmap APIs (CopyToBuffer, CopyFromBuffer, CreateCopyFromBuffer) operate on IBuffer.  I'm trying to find the right internal people to find out why that is.

To avoid excessive working set, would an option for your app be to reuse a single IBuffer instance, keeping this around from one operation to the next rather than allocating a new one each time and relying on GC to (hopefully) clean them up at some point?

I hope these questions don't seem too nitpicky, btw.  I really appreciate the work you've done to put this change together.  But, there is a cost to every new feature (API complexity, DLL size) - so we want to make sure there is compelling purpose for everything we decide to add, with no other reasonable ways to achieve the same ends.

shawnhar:
Hi,

After talking this over within the team, we’ve decided this isn’t an API we want to include in Win2D.  We don’t think the IMemoryBufferReference overloads add enough value over the existing IBuffer support to justify the extra size and API complexity.

For interop between APIs that use IBuffer and ones that use IMemoryBufferReference, it doesn’t seem very scalable to always provide two alternative versions of every such API.  Instead, I think a better solution would be to implement an adapter that could wrap an IMemoryBufferReference around an existing IBuffer (or vice versa).  It looks like that kind of interop could be done in either direction efficiently without copying the buffer contents, and once implemented, such an adapter could then be reused with any buffer related APIs.

With deepest apologies,

eadordzhiev:
My bad, I should have discussed it first :) And I deeply sorry for your time spent on my pull request.

The wrapper is a better idea. Should we expect it in Win2D?

Thanks in advance!

shawnhar:
Please, don't be sorry about anything!  I really appreciate all the work you put into this change - just sorry we weren't able to take it.

I don't think this buffer adapter belongs in Win2D, as it's more general than this particular API.  I have raised the issue with the core team that owns these two buffer interfaces.

Issue #292, opened 2016-03-04
Changing printers does not update the print preview


mtaron:
Hi Win2D!

Thanks for this project and all the great samples!

Just reporting a small bug in the Printing example. If you switch printers the print preview doesn't update until you change another property like Orientation. This is noticeable when switch from something like Print to PDF which has no margins to a physical printer that does.

This could be fixed by adding the following code to [PrintingExample.xaml.cs](https://github.com/Microsoft/Win2D/blob/master/samples/ExampleGallery/Shared/PrintingExample.xaml.cs):

```C#
private void PrintingExample_PrintTaskRequested(PrintManager sender, PrintTaskRequestedEventArgs args)
{
    [...]
    detailedOptions.OptionChanged += PrintDetailedOptions_OptionChanged;
}

private void PrintDetailedOptions_OptionChanged(PrintTaskOptionDetails sender, PrintTaskOptionChangedEventArgs args)
{
   if (args.OptionId == null)
   {
      // Invalidate the preview when switching printers.
      this.printDocument.InvalidatePreview();
   }
}
```

I tried to send a pull request to do so, but it said I don't have permission.

Thanks!
Michael

shawnhar:
I'm surprised that a pull request didn't work for you - we should be all set up for that.  Could you share more info about how you tried to do this and where the error appears?

mtaron:
I cloned the repo using Git Desktop, made a local branch, committed this change, and tried to publish the branch (or make a pull request) from Git Desktop. I get this error:

> Sync failed to push local changes
> It seems you do not have permission to push your changes to this repository.

Same thing happens when I try to publish from Visual Studio:

> Error encountered while publishing branch to the remote repository: Response status code does not indicate success: 403 (Forbidden).

I might have set my account up incorrectly or something?

shawnhar:
Ah, I understand now.  You were trying to push your changes directly to the Win2D repository, but that's not how github pull requests work.  The workflow is to create your own fork of the Win2D repository on github, push your changes there, and then create a pull request from your fork back to Win2D.

Docs covering this in more detail:
https://help.github.com/articles/fork-a-repo/
https://help.github.com/articles/using-pull-requests/

mtaron:
Doh! Thanks for your patience. I'm clearly new to GitHub. :| I sent a pull request.

PR #294, opened 2016-03-04
Update print preview when switching printers

mtaron:
If you switch printers the print preview doesn't update until you change
another property like Orientation. This is noticeable when switch from
something like print to PDF which has no margins to a physical printer
that does. (bug #292)

Fix is to listen to OptionChanged and invalidate the preview when an
event with a null OptionId is raised.


azurecla:
Hi __@mtaron__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

clandrew:
I took a look, as well. Looks good to me. Good catch!

mtaron:
Thanks! I don't have write access so one of you guys will need to merge the pull request.

PR #332, opened 2016-04-22
Added support for input of type System.Double

ratishphilip:
Classes **Vector2**, **Vector3**, **Vector4**, **Matrix3x2**, **Matrix4x4**, **Plane** &
**Quaternion** will also accept input of type **System.Double** along with **float**.

### DIFF
--- a/numerics/DotNet/Matrix3x2.cs
+++ b/numerics/DotNet/Matrix3x2.cs
@@ -74,6 +74,16 @@ namespace System.Numerics
             M21 = m21; M22 = m22;
             M31 = m31; M32 = m32;
         }
+
+
+        public Matrix3x2(double m11, double m12,
+                         double m21, double m22,
+                         double m31, double m32)
+        {
+            M11 = m11.ToFloat(); M12 = m12.ToFloat();
+            M21 = m21.ToFloat(); M22 = m22.ToFloat();
+            M31 = m31.ToFloat(); M32 = m32.ToFloat();
+        }
         
         
         public static Matrix3x2 CreateTranslation(Vector2 position)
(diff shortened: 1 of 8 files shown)


azurecla:
Hi __@ratishphilip__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

shawnhar:
Hello,

This numerics code in Win2D is actually only a clone of the System.Numerics.Vectors assembly which is part of .NET CoreFX.  That master code lives over at https://github.com/dotnet/corefx/tree/master/src/System.Numerics.Vectors, and would be the right place to propose this change.

Win2D directly uses System.Numerics.Vectors when building for UWP (Win10+).  It only includes this local copy of the numerics code to support older operating system (Windows 8.1 and Phone 8.1).

I don't think we'll want to take any new features into the Win2D copy of the code unless they were first accepted into the master System.Numerics.Vectors version.

shawnhar:
I should add that I'm not 100% convinced it's a good idea to provide methods that silently reduce precision.   C# is generally pretty consistent that any time you convert a larger or more precise value to something with fewer bits, and explicit cast should be required.

But, the main System.Numerics.Vectors github project is the right place to have this discussion.

ratishphilip:
Thanks Shawn. I shall ask this question in the  main System.Numerics.Vectors github project.

shawnhar:
Based on the resolution to https://github.com/dotnet/corefx/issues/8007, I think this is no longer an active request so I'm going to close the pull request.

PR #406, opened 2016-08-09
Fixed #373: Added Measure() call to the canvas control.

### LINKED ISSUE #373: UserControl an a ARM Device
Hi,

I implemented a UserControl with a Canvas on it (running in a UWP-Application). It works fine on the desktop (x86) and on a mobile emulator. But when I deploy it to a actual ARM device it just does not render anything.
Your demo application (glow text) work perfectly on my device. To show the problem I implemented a demo:

```csharp
	public class RedControl : UserControl
	{
		public RedControl()
		{
			Loaded += OnLayerLoaded;
			Unloaded += OnLayerUnloaded;
		}

		public CanvasControl Canvas { get; private set; }

		protected virtual void OnLayerLoaded(object sender, RoutedEventArgs e)
		{
			Canvas = new CanvasControl();

			Canvas.Draw += OnDraw;
			Content = Canvas;
		}

		private void OnDraw(CanvasControl sender, CanvasDrawEventArgs args)
		{
			args.DrawingSession.Clear(Colors.Red);
		}

		private void OnLayerUnloaded(object sender, RoutedEventArgs e)
		{
			// Explicitly remove references to allow the Win2D controls to get garbage collected
			if (Canvas != null)
			{
				Canvas.RemoveFromVisualTree();
				Canvas = null;
			}
		}

		protected override Size MeasureOverride(Size availableSize)
		{
			// CanvasTextLayout cannot cope with infinite sizes, so we change
			// infinite to some-large-value.
			if (double.IsInfinity(availableSize.Width))
				availableSize.Width = 6000;

			if (double.IsInfinity(availableSize.Height))
				availableSize.Height = 6000;

			return availableSize;
		}
	}
}
```
If I include this control into an UWP-App it starts up and show a red background. If I run this on my device the background stays white. The debugger shows that the Clear line is executed, but has somehow no effect.

Can anybody tell me what the problem is? Can anybody reproduce this?

BTW: This happens with the current nuget (18) Package as well as with the current source.

Thanks for your help!

mjeanrichard:
Added a call to ```canvas.Measure()``` to the GlowTextCustomControl. As discussed in Issue #373 .

### DIFF
--- a/samples/ExampleGallery/Shared/GlowTextCustomControl.cs
+++ b/samples/ExampleGallery/Shared/GlowTextCustomControl.cs
@@ -121,7 +121,7 @@ namespace ExampleGallery
             {
                 instance.canvas.Invalidate();
                 instance.InvalidateMeasure();
-            }            
+            }
         }
 
         // This is the amount that we grow the desired size by (to account for the glow)
@@ -142,7 +142,11 @@ namespace ExampleGallery
             var layout = CreateTextLayout(device, availableSize);
             var bounds = layout.LayoutBounds;
 
-            return new Size(Math.Min(availableSize.Width, bounds.Width + ExpandAmount), Math.Min(availableSize.Height, bounds.Height + ExpandAmount));
+            Size desiredSize = new Size(Math.Min(availableSize.Width, bounds.Width + ExpandAmount), Math.Min(availableSize.Height, bounds.Height + ExpandAmount));
+            if (canvas != null)
+                canvas.Measure(desiredSize);
+
+            return desiredSize;
         }
 
         private void OnDraw(CanvasControl sender, CanvasDrawEventArgs args)


azurecla:
Hi __@mjeanrichard__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

samples/ExampleGallery/Shared/GlowTextCustomControl.cs
```diff
@@ -142,7 +142,10 @@ protected override Size MeasureOverride(Size availableSize)
             var layout = CreateTextLayout(device, availableSize);
             var bounds = layout.LayoutBounds;
 
-            return new Size(Math.Min(availableSize.Width, bounds.Width + ExpandAmount), Math.Min(availableSize.Height, bounds.Height + ExpandAmount));
+            Size desiredSize = new Size(Math.Min(availableSize.Width, bounds.Width + ExpandAmount), Math.Min(availableSize.Height, bounds.Height + ExpandAmount));
+            canvas?.Measure(desiredSize);
```
shawnhar: The ?. operator was a new addition in C# 6.0, but Win2D still supports Visual Studio 2013 so I'm afraid you cannot use this :-(

mjeanrichard: Fixed. I am already so used to these new features that I don't even notice them anymore.

clandrew:
Between issue #373 and reading MSDN documentation for MeasureOverride, calling Measure on child XAML elements is the right thing here.
I see that you fixed the issue where we're depending on VS2015-specific compiler features.
This change looks good to me.

Issue #409, opened 2016-08-15
CanvasTextFormat.GetSystemFontFamilies() returns Fonts that cause apps to crash


mscherotter:
If a user installs fonts from Adobe Creative Cloud Fonts, they are listed in the CanvasTextFormat.GetSystemFontFamilies() list.  If the fonts are used in a UWP app then the UWP app crashes.
`<TextBlock Text="Tekton Pro Test" FontFamily="Tekton Pro"/>`
CanvasTextFormat.GetSystemFontFamilies() should only return fonts that can be rendered in UWP apps without crashing.
Repro steps:
1. Install an Adobe Creative Cloud Font 
2. Run the Win2D Gallery App
3. Navigate to the TextLayout demo
4. In the Hamburger menu, select the Font combo and scroll to the font that was installed - CRASH!

clandrew:
Hi, thanks for sending along this report! GetSystemFontFamilies should be returning usable fonts, so it seems possible that there's a real bug here.
I'll take a look at this.

shawnhar:
The root cause of this is broader than just Win2D, as any use of that font causes issues with other apps too (eg. XAML).  The DWrite team is investigating.

For my reference, this is tracked by internal bug #8512526.

shawnhar:
Update: this issue is understood, and it will be fixed in the next version of Windows.

shawnhar:
This fix is available in the most recent update to Windows 10.

PR #410, opened 2016-08-15
Fixed issue where certain fonts intstalled by 3rd party apps can't be used in UWP and are returned by CanvasTextFormat.GetSystemFontFamilies()

mscherotter:
Certain fonts installed by third party libraries were enumerated by CanvasTextFormat.GetSystemFontFamilies() but when used they were causing UWP apps to crash.  See #409 for a repro.  This code checks to see if all fonts can be created before returning.  It does not have any noticeable performance penalty - tested with 176 fonts installed.

### DIFF
--- a/winrt/lib/text/CanvasTextFormat.cpp
+++ b/winrt/lib/text/CanvasTextFormat.cpp
@@ -6,6 +6,7 @@
 
 #include "CanvasTextFormat.h"
 #include "TextUtilities.h"
+#include <vector>
 
 using namespace ABI::Microsoft::Graphics::Canvas::Text;
 using namespace ABI::Windows::Storage;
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@mscherotter__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

clandrew:
Thanks for sending along this pull request. I'll take a look at this, in conjunction with #409 to get a sense of what's going on.

clandrew:
We've had the opportunity to understand this issue better. There's an external (non-Win2D) issue preventing Creative Cloud fonts from being used with UWP. We're tracking this issue internally.

I am not convinced we should take this change, because it has very expensive performance implications while failing to address the root problem. 

The performance hit of instantiating every font on the system may not have been evident on the environment in which you tested it, but on some environments it's prohibitively slow. 

In particular, instantiating all fonts can be noticeably slow on Windows phone hardware. And a system (both Windows and Windows Phone) can have far, far more fonts installed than just the in-box ones. This performance hit of this change would scale with that number of fonts. I don't think it's acceptable to introduce such a significant performance hit for all apps that do something as simple as enumerate system fonts (e.g., by adding a font picker).

The problem of a non-instantiatable font isn't specific to this situation. Generally speaking, a system could have any number of invalid, or malformed fonts, for any number of different reasons. Because of how costly it is to sanitize them out at font-enumeration time, it may make more sense to handle the error instead when the font is used.

So, I will not be taking this PR. I really appreciate your attention to this font installation issue, and look forward to seeing it get resolved. We'll be sure to follow up with you via issue #409.

PR #581, opened 2017-11-22
Win2D geometry interop with RS4 Windows.UI.Composition.CompositionPath

adadi-MS:
Tested, the versioning needs work.  This is just and FYI


adadi-MS:
Awesome, thanks Shawn!

Issue #628, opened 2018-07-17
CanvasSvgDocument FindElementById method throws System.AccessViolationException


mobilefringe:
When trying to find an element by its ID in a CanvasSvgDocument, it works if the ID exists in the SVG document. However, if you try looking for an ID that does not exist, a System.AccessViolationException will be thrown with the following error message: System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'

This is occurring for me on Visual Studio 2017 on a project targeting Windows 10, version 1803 with Win2D 1.23.0.

clandrew:
Thank you for the report. I'll take a look into reproducing the issue on our end.

clandrew:
I reproduce the issue. Fix is in progress.

clandrew:
Fix has been merged with PR #635.
Thank you again for reporting this! I'm closing the issue, seeing no further action required for this, but feel free to get in touch if you need anything else.

PR #635, opened 2018-09-19
Fix AV when a non-existent element is passed to CanvasSvgDocument.FindElementByID

clandrew:
This is for fixing issue #628 - "CanvasSvgDocument FindElementById method throws System.AccessViolationException".


damyanp:
Looks good to me!

PR #655, opened 2018-11-17
Add C++/WinRT example to interop article

benstevens48:
This is an initial attempt to explain how to use C++/WinRT to interop between Win2D and Direct2D. There are probably a few typos etc. I wasn't sure how to view the aml file in its final state so I can't tell how it looks.

Probably someone should test the code to make sure nothing got broken in the process of putting into the documentation. Also, I should add that I am very new to C++/WinRT and in fact I'm not very familiar with C++ in general, so I may have used bad practice and the wrong terminology in various places. Hopefully someone can point that out if so.

I hope this is helpful.

### DIFF
--- a/winrt/docsrc/Interop.aml
+++ b/winrt/docsrc/Interop.aml
@@ -345,6 +345,65 @@ Licensed under the MIT License. See LICENSE.txt in the project root for license 
         </code>
       </content>
     </section>
+
+    <section>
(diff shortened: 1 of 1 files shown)


shawnhar:
Many thanks for the contribution!  This looks good to me, and it built successfully, so I'm going to go ahead and merge it in.

PR #662, opened 2018-12-22
Update README.md for C++/WinRT

Berrysoft:
C++/WinRT is a recommended replacement for C++/CX, and can get use of Win2D without difficulty.


shawnhar:
Thanks for the contribution!

PR #665, opened 2019-01-15
ARM64 support

fobrs:
added ARM64 support. Need still some work: In example gallery I needed to remove the LiveTileUpdater background task from the manifest because of errors I couldn't resolve otherwise. The min OS version is now 1803.


clandrew:
This is awesome!! Thank you for doing this.

shawnhar:
+1 on thanks for doing this!

I have one question about the change: Win2D.cpp.props changes WindowsTargetPlatformVersion to 10.0.17763.0 while leaving WindowsTargetPlatformMinVersion at 10.0.10240.0, but the C# equivalent, Win2D.cs.props, changes TargetPlatformVersion to 10.0.17763.0 and TargetPlatformMinVersion to 10.0.17134.0.  Why can't that last be left at 10.0.10240.0, as in the .cpp.props?  I think that would be the key for maintaining support for using Win2D on older Windows versions.

Related note: a number of the .vcxproj files add a ProjectConfigurations element, plus defines for WindowsTargetPlatformVersion and WindowsTargetPlatformMinVersion.  These should not be needed, as those things are pulled in from a single source of truth in Win2D.cpp.props.  It would be good to maintain that design so the .cpp.props is the only place that needs to be changed when moving forward to even newer SDK versions in the future.

Thanks!

fobrs:
I will look if I get it compiled with the changes you propose. What is your opinion about the added <PackageReference> tags in the *.csproj file?

shawnhar:
> What is your opinion about the added tags in the *.csproj file?

Good point - we shouldn't need target platform version information in the .csproj files either, actually.

fobrs:
Ok, sorry, but I actually meant the added <PackageReference> tags in the .csproj files. I used them to get the version 2.2.1 of the .net native compiler working. Could there be another way to accomplish this?

fobrs:
Ow, the editor was eating my `<PackageReference/>` strings

shawnhar:
I'm not following which tags you mean?  Looking at CompositionExample.csproj for instance, the only added tags are:

    <TargetPlatformVersion>10.0.17763.0</TargetPlatformVersion>
    <TargetPlatformMinVersion>10.0.10240.0</TargetPlatformMinVersion>

which I think should not be needed here (all this info should come from the .cs.props include instead)

fobrs:
They are in ExampleGallery.BackgroundTask.uap.csproj , ExampleGallery.Effects.csproj  and ExampleGallery.uap.csproj. Added in the last commit of the pull request (to get the ExampleGallery compiling succesfull).

shawnhar:
Ah, I see the <PackageReference> additions for the .NET native compiler.  That seems reasonable to me.

shawnhar:
Checking in on the status of this PR - looks like there are three things that need to be addressed before it can be merged:

1. Avoid changing the min platform version, so the resulting Win2D binaries (at least for configurations other than ARM64) can still be used on older Windows versions.

2. Get BackgroundTask.LiveTileUpdaterTask added back to the Example Gallery manifest.

3. Remove unnecessary &lt;ProjectConfigurations&gt; elements from the  .vcxproj files.

Is that the complete set?

Thanks!

fobrs:
1 and 3 are trivial and can be done in my PR.

2 needs more work.

fobrs:
1. When I set the min platform version less than 16299, ARM64 build is not successful
2. Adding back BackgroundTask.LiveTileUpdaterTask to the  Example Gallery manifest. leads to the error: d:\tools\win2d\bin\uaparm64\release\examplegallery.uap\AppxManifest.xml : error APPX0501: Validation error. error 80080204: App manifest validation error: Line 39, Column 12, Reason: If it is not an audio background task, it is not allowed to have EntryPoint="ExampleGallery.BackgroundTask.LiveTileUpdaterTask" without ActivatableClassId in windows.activatableClass.inProcessServer.

I've no time to debug these problems... Maybe you could try it yourself, as you do not need an ARM64 device to see those problems....

build/Win2D.cpp.props
```diff
@@ -30,6 +34,10 @@
       <Configuration>Release</Configuration>
       <Platform>ARM</Platform>
     </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|ARM63" Condition="'$(ApplicationType)' != ''">
```
kvpt: Typo here => ARM63

fobrs: Good find

fobrs:
We still have the problem about entering a min version of 16299. If this is not specified the ARM64 build will fail. Isn't it possible to fork Win2D for newer (> 16299) win10 SDK's?

MagicAndre1981:
why not set 16299? VS2019 drops all older [UWP SDKs in designer](https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-preview#universal-windows) 

> XAML design time tooling for UWP apps targeting Windows 10 SDKs before the Windows 10 Fall Creators Update (build 16299) has been removed. Retarget your applications to Windows 10 version 1709 or higher to use the XAML Designer, or use the XAML editor.

Build 16299 reaches End of support next month for Home/Pro and for Enterprise/Education in Fall 2019. So set it to 16299 and eveything is fine.

fobrs:
shawnhar should decide...

MagicAndre1981:
btw, 1.23.0 already requires 17134

winrt/dll/winrt.dll.uap.vcxproj
```diff
@@ -13,11 +47,11 @@
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
     <ItemsRootNamespace>Microsoft.Graphics.Canvas</ItemsRootNamespace>
     <TargetName>$(ItemsRootNamespace)</TargetName>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.17763.0</WindowsTargetPlatformMinVersion>
```
lhak: The platform versions are already defined in the .props file so they should not be redefined in all the projects files.

lhak:
Would it be possible to add a condition to the the .props file so that the arm64 build sets the minimum version to 16299 but the other builds remain at the current value?

fobrs:
> btw, 1.23.0 already requires 17134

I don't think so. In the original files is specified:

`<WindowsTargetPlatformMinVersion>10.0.10240.0</WindowsTargetPlatformMinVersion>`

MagicAndre1981:
Win2D [needs 17134](https://github.com/Microsoft/Win2D/blob/master/build/nuget/Win2D.common.targets#L5) as MinVersion:

```
<PropertyGroup>
    <Win2DMinPlatformVersion>10.0.17134.0</Win2DMinPlatformVersion>
  </PropertyGroup>

```

so I needed to add `<DisableWin2DPlatformCheck>true</DisableWin2DPlatformCheck>` to my csproj to skip the test and use it in 16299.

shawnhar:
There are three different SDK versions relevant to Win2D:

1) The version needed to build Win2D itself.  This is typically the latest official (non-prerelease) version of Visual Studio.

2) The version needed to build apps that use Win2D via its precompiled NuGet package.  This is usually the same as the first version, but it does not have to be (at some points in time Win2D has required a newer version to build itself, but can still be used by developers building apps with some older version).

3) Finally, there's the minimum version of Windows that apps created using Win2D can run on.  This is the WindowsTargetPlatformMinVersion, and it must never be increased past 10.0.10240.0 because we do not want to force all Win2D apps to require a newer OS.

ARM64 adds a complexity because there was no such thing as ARM64 in 10240.  The right thing is probably to give ARM64 a different min version than the other Win2D architectures.  That way ARM64 apps will require a newer OS (which is fine because there are no older Windows builds running on ARM64) but customers using other architectures do not lose any access to older OSes.

MagicAndre1981:
@shawnhar 

Xamarin.Forms 3.6 adds Win2D as dependency, so I installed last version via nuget and here I had to add `<DisableWin2DPlatformCheck>true</DisableWin2DPlatformCheck>` to my csproj file otherwise I was not able to compile my app.

fobrs:
I've just **successfully** compiled and run a Win2D test app, based on [this tutorial](https://microsoft.github.io/Win2D/html/QuickStart.htm), with VS2017 15.9.9.

I used Win2D 1.23 from nuget.

I've this in my csproj:

`<TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.17134.0</TargetPlatformVersion>
<TargetPlatformMinVersion>10.0.10240.0</TargetPlatformMinVersion>`

build/Win2D.cpp.props
```diff
@@ -45,7 +53,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
jkunkee: WindowsTargetPlatformVersion is usually a project-wide setting, and it often has follow-on effects. Typically it is changed by itself all at once and the change followed by extensive testing. I recommend this because it allows SDK version differences to be dealt with separate from architecture differences.

In this case, the README states that it is 17134 and Win2D itself is built for 17134:
    build\nuget\Win2D.common.targets:5:    <Win2DMinPlatformVersion>10.0.17134.0</Win2DMinPlatformVersion>

The way to build without changing the SDK version is to install the older SDK using the "Individual Components" pane of the Visual Studio Installer.

azchohfi:
Is anything blocking this from being merged?

shawnhar:
> Is anything blocking this from being merged?

There are a couple of issues with this PR: it breaks Win2D compatibility with older Windows versions, and it breaks the Example Gallery due to missing BackgroundTask.LiveTileUpdaterTask.

https://github.com/microsoft/Win2D/pull/695 is a parallel effort to do the same thing, which seems closer but got stuck while attempting to validate the changes.

shawnhar:
We ended up taking a different approach to ARM64 support via https://github.com/microsoft/Win2D/pull/722, so this PR is no longer needed.   Thanks for your work on it!

PR #670, opened 2019-02-07
A sample for using Win2D in XAML island on .NET Core 3 

devsko:
Also adding support for VS + msbuild 2019 and fixing some build and test problems.


devsko:
cc @jasonwurzel @pcfulife @jtorjo

jtorjo:
@devsko Awesome, will try in a bit! I'm assuming it'll work only on .Net Core 3, not on any previous .net code, right?

shawnhar:
This is fantastic, thanks for putting it together!

I should note that the change not only adds support for VS 2019, it looks like it _requires_ 2019.  This is fine, as Win2D typically requires the latest official release of Visual Studio to build it (although it must be possible to build client apps using Win2D with older Visual Studio versions).  We don't generally bother doing the extra engineering work to make Win2D itself compatible with more than one VS version at the same time.

However, since VS 2019 is still in preview, we should hold off on merging this PR to master until after VS 2019 is officially released.  The Win2D master branch should not depend on a prerelease VS version.

samples/XamlIslandSample/Microsoft.Toolkit.Wpf.UI.XamlHost/Microsoft.Toolkit.Wpf.UI.XamlHost.csproj
```diff
@@ -0,0 +1,49 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
+    <RootNamespace>Microsoft.Toolkit.Wpf.UI.XamlHost</RootNamespace>
+    <AssemblyName>Microsoft.Toolkit.Wpf.UI.XamlHost</AssemblyName>
+
+    <Title>Windows Community Toolkit XAMLHost for WPF</Title>
+    <Description>This library provides XAML islands helpers for WPF. It is part of the Windows Community Toolkit.</Description>
```
shawnhar: Is it really necessary to duplicate this code here rather than just pulling the toolkit down from NuGet?

tools/checkcode/RunCodeChecks.proj
```diff
@@ -31,8 +31,10 @@
     <!-- File extensions and directories to check vs. ignore -->
     <CppExt>cpp;h;inl;cs;idl;rc;hlsl</CppExt>
     <XmlExt>xml;xaml;aml</XmlExt>
-    <SkipExt>;gitattributes;gitignore;orig;jpg;png;dds;txt;md;pfx;cmd;bat;proj;nuspec;props;targets;def;sln;csproj;vcxproj;vcxitems;filters;projitems;shproj;shfbproj;content;config;appxmanifest;suo;ipch;sdf;log;opensdf;user;exe;bmp;ico;tif;vsdx;svg;json;bin;opendb;db</SkipExt>
+    <SkipExt>;gitattributes;gitignore;orig;jpg;png;dds;txt;md;pfx;cmd;bat;proj;nuspec;props;targets;def;sln;csproj;wapproj;vcxproj;vcxitems;filters;projitems;shproj;shfbproj;content;config;appxmanifest;suo;ipch;sdf;log;opensdf;user;exe;bmp;ico;tif;vsdx;svg;json;bin;opendb;db</SkipExt>
     <SkipDir>.git;obj;bin;TestResults;build;.nuget;packages;.vs</SkipDir>
+    <!-- The temporarily included files from Microsoft.Toolkit have their own license text -->
```
shawnhar: Why temporarily?   (related to earlier comment about whether this can be pulled from NuGet)

winrt/test.managed/PixelShaderEffectTests.cs
```diff
@@ -1216,13 +1216,13 @@ public void PixelShaderEffect_InputRectTooBigError()
...
 
-                Utils.AssertThrowsException<Exception>(() =>
+                Assert.ThrowsException<Exception>(() =>
                 {
                     using (var ds = renderTarget.CreateDrawingSession())
                     {
                         ds.DrawImage(effect);
                     }
-                }, " graph could not be rendered with the context's current tiling settings. (Exception from HRESULT: 0x88990027)");
+                });
```
shawnhar: I assume this is also a fix for non-en culture?   Can the test still look for presence of 0x88990027 in the exception string?  Without that, it's not really testing for the specific failure condition that is expected any more.

winrt/test.nativecomponent/winrt.test.nativecomponent.uap.vcxproj
```diff
@@ -34,6 +34,7 @@
       <AdditionalIncludeDirectories>..\Inc;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>_WINRT_DLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <WarningLevel>Level4</WarningLevel>
+      <TreatWarningAsError>false</TreatWarningAsError>
```
shawnhar: Why?   Policy is for Win2D to build without warnings...

devsko:
@jtorjo I only tried .NET Core

pcfulife:
@devsko Wow! Thanks a million! I will try it!

devsko:
@shawnhar You are right this PR requires VS 2019. (I only tried to create a .NET Core 3 project with VS 2017 once). I did this sample mainly to make it possible to find and fix issues with Win2D / XAML islands (we need this in our closed source project soon). Maybe it is a good idea to have a preview branch for this PR and the coming fixes. When VS / .NET Core / Windows xxx gets released it can be merged to master.

samples/XamlIslandSample/Microsoft.Toolkit.Wpf.UI.XamlHost/Microsoft.Toolkit.Wpf.UI.XamlHost.csproj
devsko: Problem is that all official preview builds that support .NET Core 3 only work on Windows insider builds. On RS5 everything builds and starts but crashes with TypeLoadException when the app window is moved. I had a PR over there but didn't get any response. I will change to the latest preview package from myget and update the readme

tools/checkcode/RunCodeChecks.proj
devsko: will change

winrt/test.managed/PixelShaderEffectTests.cs
devsko: Yes, on my box the message was very different. I was thinking the same. Will add a check for HResult

winrt/test.nativecomponent/winrt.test.nativecomponent.uap.vcxproj
devsko: Without this the command line build failed with C4244 possible loss of data (in VS it builds without problem) [here](https://github.com/Microsoft/Win2D/blob/d0ef1aa3a93045b910fa67f31dddf1b41fdf70d1/winrt/test.nativecomponent/ShaderCompiler.cpp#L16)
My C/C++ skills are very basic so I did it this way. There are better ways for sure. Do you have a hint?

pcfulife:
It doesn't throw COMException anymore, thanks. But, it throws FileNotFoundException when I create CanvasAnimatedControl. I think that I need to copy module file (Microsoft.Graphics.Canvas.dll). But, it already exists in (WAP project directory)\bin\x64\Debug\AppX\(Original project name).

devsko:
Do you use the sample project or did you create your own? I listed all necessary steps in readme file. Especially referencing the VC Libs in package manifest is important. Btw: I found the not found files with sysinternals process monitor.

pcfulife:
@devsko Oh, I am sorry. I missed that. I will read that carefully.

winrt/test.nativecomponent/winrt.test.nativecomponent.uap.vcxproj
shawnhar: Ah, I see what's happening there.  Looks like the newer compiler is smart enough to notice that the sloppy string conversion in that test code is indeed sloppy :-)

I haven't tested this, but something like the following should work (to replace the first std:;string constructor - repeat same pattern for each of the three strings):

    std::string asciiShader(shaderCode->Length(), ' ');
    std::transform(shaderCode->Begin(), shaderCode->End(), asciiShader.begin(), [](wchar_t c) { return static_cast<char>(c); });

clandrew:
Overall +1, this change looks good to me! Huge kudos for getting this working!

The requirement of VS2019 means that we shouldn't merge this to master until 2019 is RTMed. But, we could put it in some official kind of branch (i.e., something called *xaml_islands* or so forth) until we can merge it to master and take a dependency on VS2019.

winrt/test.nativecomponent/winrt.test.nativecomponent.uap.vcxproj
devsko: It works. Thanks

devsko:
@clandrew Sounds good. Can you create a branch please. I get a 403 :) 
I will then move this PR to the new branch.
Next step is a updated nuget package with netcoreapp target and a fix for CanvasControl

clandrew:
@devsko Sure thing. I created the branch xaml_islands off of current master for you.

TaeNet:
I have made a 2D game since 2017.12 by using WPF. I know that WPF is not good for game. But, I intended it as a small and simple game.
However, it grows bigger than I could have expected. WPF cannot sustain stable FPS so that the frame of my game is so fluctuate. I was so depressed because I spent almost 1 year to this project.
But after I know that Win2D will be able to use in WPF with .NET Core 3.0, I am so happy to improve my game. Thanks @devsko, all Win2D and .NET Core team!

shawnhar:
Thank you so much for putting this together and incorporating all the feedback!   I'm completing the PR into the xaml_islands branch, where we'll leave it staged until this technology moves out of preview state.

Issue #600, opened 2018-05-03
Tracking item: remove Win2D workarounds for vstest bugs


shawnhar:
Win2D currently contains workarounds for these bugs, which should be reverted once the fixes are available in our build environment:

https://github.com/Microsoft/vstest/issues/1533
https://github.com/Microsoft/vstest/issues/1534
https://github.com/Microsoft/vstest/issues/1535

PR #674, opened 2019-02-15
Include XamlIslandSample in Win2D.proj

devsko:
This is gone a bit wild...
- Added Win32 configuration to XamlIslandSample projects
- Changed explicit imports of Win2D.cpp/cs.props/targets to implicit Directory.Build.props/targets to support .NET Core SDK projects
- Introduced __TargetPlatformVersion/__TargetPlatformMinVersion in Directory.Build.props as the one and only place for Windows versions
- Tweaking some backslashes in paths
- Fixing #600 - Verified that no more workarounds needed for
  - IntermediateOutputPath in test projects
  - Run of C++ UWP tests during command line build
  - NuGet restore


shawnhar:
This looks good to me.  Thanks for putting it together!

I'd appreciate a sanity check from @damyanp before merging, as he originally implemented most of the build files this is touching.

lhak:
Would it be possible to add a configuration so that the nuget package can be installed into a .net framework application?

devsko:
@lhak That is in the pipeline but it takes time. First want to fix .NET Core scenario
But it should not be that difficult

damyanp:
LGTM.  TIL about Directory.Build.* :)

PR #675, opened 2019-02-18
Add a method to CanvasControl to manually invoke the rendering event

lhak:
CanvasControl currently does not work in XAML islands. The issue is that the Draw event is never called because of the following reasons (seen on Windows 10 1809 and the current insider fast build):

*The CompositionTarget.Rendering event is never called
*The associated window will always be reported as not visible

The PR adds a new method to the CanvasControl which is basically a copy of the method invoked by the rendering event, but does not check for window visibility. In my tests manually calling this method from the CompositionTarget.Rendering event (the .net version in System.Media) results in a working CanvasControl object.


pcfulife:
When I use Win2D in UWP, calling in Environment.CurrentManagedThreadId returns 3. But, when do it in CompositionTarget.Rendering (Both in WPF's and UWP's one), it returns 1.
So, draw event doesn't need to run in UI thread.

Anyway, I am using CanvasSwapChainPanel in WPF. A loop is needed to animate CanvasSwapChainPanel. I found that framerate is lower and unstable when I place loop in CompositionTarget, but if in new thread (Task.Run), its framerate is much better than CompositionTarget.

lhak:
In my tests, the Draw event is actually called on the UI thread, but its number can differ from 1 for UWP. The Rendering events also seems to behave differently for UWP and WPF/Winforms. For UWP, it is called every 16ms (=60Hz refresh rate), but it fluctuates around 30-45ms for WPF/Winforms. This should still be sufficient for CanvasControl, because it is more suitable for static content according to the documentation.

shawnhar:
Thanks for putting this together!

Couple of suggestions:

* Can you pull the existing drawing code out into a helper function so it can be reused rather than duplicated?  There's enough logic in there to be worth avoiding two copies of it.

* Regarding naming, I think this should be just ManualRender rather than ManualRenderTick.   Would appreciate input from @clandrew and @damyanp on this.

Thanks!

btw. the numeric value of a thread ID is not significant - that's just an identifier/handle.

devsko:
@lhak What do you think about adding a public setter for the `BaseControl.m_isVisible` flag instead. All the event handling would stay internal to the control and the refresh rate wouldn't change.
I was thinking about a fix like 
- `bool BaseControl.EnableWindowVisibilityDetection(bool)` when set to false stops listening for visibility change of the window and sets m_visible to true
- `BaseControl.set_IsVisible(bool)` sets the flag explicitly that starts/stops rendering

One could think about stopping the listener implicitly when set_IsVisible() is called but that is maybe too much magic.

shawnhar:
> What do you think about adding a public setter for the BaseControl.m_isVisible flag instead.

Mmm, that's a good idea!   So the composition rendering event still works in XAML islands as long as the visibility detection is overridden?

devsko:
The whole composition rendering event would stay inside CanvasControl. Only the automatic detection of changed window visibility could be stopped and could then be done from outside - calling IsVisible = true when the WPF window gets visible and vice versa

devsko:
>So the composition rendering event still works in XAML islands as long as the visibility detection is overridden?
According to my tests, yes. I tested with 17763 (NOT insider previews). I set a breakpoint and changed the BaseControl.m_Visible to true. All events were fired.

lhak:
Ok, seems that I was wrong and the UWP composition event does actually work on the current released Windows version and insider preview. So overriding the visibility property is enough to unblock CanvasControl (and CanvasAnimatedControl too!) and the approach proposed by @devsko should be implemented. Do you think we need a way to reenable the visibility autodetection or is it sufficient to add a single method (SetVisible(bool visible)?) that always disables the autodetection?

devsko:
@lhak that sounds good
>Do you think we need a way to reenable the visibility autodetection

I don't think so

winrt/lib/xaml/BaseControl.h
```diff
@@ -502,9 +506,19 @@ namespace ABI { namespace Microsoft { namespace Graphics { namespace Canvas { na
...
+                return(m_isVisibleOverrideValue);
+            }
             return m_isVisible;
         }
 
+        void SetVisibleOverride(bool isVisible)
+        {
+            m_isVisibleOverrideEnabled = true;
+            m_isVisibleOverrideValue = isVisible;
+        }
```
devsko: Does this need a lock? I have no idea which state needs one and which not. Just looking at the other code.
Maybe also IsVisible() needs one because a possible race.

devsko:
I tested the current xaml-island branch including this PR, built and packed with build.cmd  in all 4 configurations 
- .NET Core WPF
- .NET Core WinForms
- .NET Framework WPF
- .NET Framework WinForms

Environment:
- Windows 17763
- VS 2019 Preview 3
- .NET Core 3.0.100-preview-010184 / .NET Framework 4.7.2
- Microsoft.Toolkit.WPF/Forms.UI.XamlHost  6.0.0-preview1

Win2D package can be added to all projects. Build and APPX packaging works (when all workarounds are applied).
CreateResources and Draw events are fired as expected in all apps (when SetVisible(true) was called)

Lots of fun :) Thank you

lhak:
I added the locking to SetVisibleOverride(). Anything else I should change?

shawnhar:
> Anything else I should change?

Other open issue is naming of the new API.   I suggest something like ForceVisibility or OverrideVisibility, because SetVisible could easily be confused with the XAML Visibility property.

devsko:
>Other open issue is naming of the new API

It is not really about visibility at all. It's about not suspending rendering as response of pretended invisibility. I am not a native speaker (and not good in naming) but maybe it is better going in the direction `DisableRenderSuspension` or so.

shawnhar:
> It is not really about visibility at all. It's about not suspending rendering as response of pretended invisibility. 

Good point.  Summoning @damyanp, who is better than me at naming...

damyanp:
Sorry coming to this a bit late.

How do we expect developers to use this?  Can you point me towards any sample code please, or maybe paste something here?

Will they always need to write code to enable this, or is there some way to hook things up through XAML?  Is there isn't, should (or can) there be?

Should this be a property?

devsko:
@damyanp Usage is like
``` c#
var canvas = new CanvasControl();
canvas.SetVisible(true); // or however the naming is in the end
```
The method/property should ONLY be called, if a CanvasControl is instantiated inside a XAML island (WPF or WinForms).
There is no way to write XAML directly inside a XAML island (at least I'm not aware of) so it is sufficient when it is accessible in code.

The semantics are that you can explicitly enable/disable rendering of the control. In both cases from this moment on the "real" visibility of the window is ignored that normally stop/start rendering.

damyanp:
An advantage of a property is that you can inspect in the debugger to see if it has been set correctly (and have confidence that the object itself still thinks it is set).

I think we imagine that apps might call it with false when they themselves determine that it's no longer visible / shouldn't redraw?  In this case it would be particularly interesting to inspect if it's forced on or off.

What if we had:

```
canvas.EnableRendering = CanvasControlRendering.Default; // (the default, UWP/XAML behavior)
canvas.EnableRendering = CanvasControlRendering.ForceOn;
canvas.EnableRendering = CanvasControlRendering.ForceOff;
```

damyanp:
Thinking a little harder - `EnableRendering` isn't a great name for a property as "Enable" is a verb rather than a noun.  Also we tend to refer to "Drawing" rather than "Rendering" in Win2D.

Some other ideas:

* `DrawMode`
* `VisibilityMode`
* `DrawingEnabledMode`

Any thoughts (or votes)?

shawnhar:
I vote for VisibilityMode.  This code seems very clear:

canvasControl.VisibilityMode = CanvasVisibilityMode.ForceOn;

damyanp:
`canvasControl.VisibilityMode = CanvasVisibilityMode.ForceVisible;`?

shawnhar:
+1

Enum options would be CanvasVisibilityMode.Default, CanvasVisibilityMode.ForceVisible, and CanvasVisibilityMode.ForceHidden?

devsko:
I can live with that but find it still a bit misleading. Needs good documentation :)

lhak:
Ok, I moved the visibility setting to a property in BaseControl which is accessible for CanvasControl, CanvasAnimatedControl and CanvasVirtualControl objects.

damyanp:
Great stuff!  This looks good to me, thank you!

shawnhar:
Agreed, I like this new API design a lot.  Only remaining issue is the IsVisible method needs a lock.

shawnhar:
Looks great.  Many thanks for doing this!

markzielinski:
Hi, do you have an ETA on when this will be pushed out to an official release and package to reduce the necessary steps, such as building from custom source?

shawnhar:
Merging this into an official Win2D release is dependent on the necessary SDK and Visual Studio bits moving from preview to released state.

lhak:
The code in the xaml_islands branch can be compiled with Visual Studio 2017 and the current SDK (it actually uses an older SDK at the moment). However, I think the next Visual Studio version and Windows SDK will be released soon, so it might make sense to wait until then. The only exception I can think of is the xaml island sample which uses .net core 3. Maybe this one can be taken out from the standard build configuration so we do not have to wait until the .net core 3 release to get a new nuget package?

PR #676, opened 2019-02-19
Adding WORKAROUNDs for VS issues to XAML island sample

devsko:
I recognized that the .deps.json file (a .NET Core thing) is not included in the appx deployment. That leads to subtle issues (e.g. the .net runtime version information doesn't work !).
Added workarounds to sample and updated README
(Visual Studio bug: https://developercommunity.visualstudio.com/content/problem/462646/referencing-native-winrt-component-in-net-core-wpf.html)


shawnhar:
Suggestion - consider including a link to the VS bug in each place that has a workaround comment, so future maintainers can easily check whether that has been fixed and identify all the workarounds that are no longer needed?

I don't feel super strongly about this, so am also happy to just go ahead and complete the PR if you don't think that is worth bothering with :-)

devsko:
I add it - we will see how long it is active :)
I will have a look at what changes in VS/.NET Core anyway and remove the workarounds or the comments

PR #687, opened 2019-03-19
Make the XAML island sample a bit more interesting

devsko:
The sample now shows the particle sample from ExampleGallery inside the XAML island.


devsko:
Ah. Didn't know this. As far as i see, the script still works but the XAML island sample is not included of course.
The output is almost in sync with the Win2d-Samples repo. The new Directory.Build.props/.targets are missing and there is a part of ExampleGallery.Direct3DInterop.uap.vcxproj missing.
I can have a look for the missing props/targets next week.
Do you know about the vcxproj and do you want the XAML island sample included?

shawnhar:
These sample export issues will need to be fixed before the XAML islands branch merges back to master, but I don't see a need to hold up this particular PR due to a preexisting condition in the branch it targets.

I'd like a second opinion on this pattern of reusing implementation files from one sample into another.  @damyanp @clandrew does that seem ok to you?

damyanp:
It does feel a little awkward to me.  I know that when I'm browsing samples in GitHub repos I'll often look at the directory view rather than open up a project file.  So I'd probably completely miss the particle code in this case and be left wondering where it is.

Is this something that the sample export could fix up by duplicating the files in the exported version?

shawnhar:
I'm pretty sure the sample export does duplicate any referenced files?

damyanp:
Ah yes, so it does!  That would suggest that this isn't the first time we've come across this question.  LGTM!

shawnhar:
Thanks Damyan - I'll go ahead and merge this then

PR #695, opened 2019-04-27
Add minimum ARM64 support

jkunkee:
This change adds ARM64 to all of the project configurations and updates CONTRIBUTING.md to mention ARM64.

This is inspired by https://github.com/Microsoft/Win2D/pull/665 but is tidier.

* `build.cmd` succeeds
* The generated Nuget package works on ARM64 when used in the [Hello World quick start](https://microsoft.github.io/Win2D/html/QuickStart.htm) project.
  * Remote debugging itself requires some shenanigans for ARM64, but the app builds, deploys, and runs correctly.
  * Note that the project needs to have a `TargetPlatformMinVersion` of 16299 or later for ARM64. This can be set directly in the vcxproj file:
```
  <PropertyGroup Condition="'$(Platform)' == 'ARM64'">
    <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
  </PropertyGroup>
```
* The sample projects do not build. I am not enough of a Nuget or UWP wizard to track down why, but it is probable one of the test frameworks is missing ARM64 support.

@shawnhar, I'm happy to expand this; see https://github.com/jkunkee/Win2D/commits/arm64-build-changes for some of my other thrashing.


shawnhar:
Hi, thanks for putting this together!   The approach looks good to me and all the changes seem correct.  I tested this out and confirmed many things are working correctly:

- build.cmd works to create a NuGet package including ARM64 binaries
- Building for x86 inside Visual Studio still works too
    - All x86 unit tests still pass
    - x86 Example Gallery works
- The ExportSample script still works, and includes ARM64 config in the exported samples

But there are a couple of things that will need addressing before this can be merged to master for a new Win2D release:

1) "msbuild /m win2d.proj" fails (trying to build test project for ARM64)
2) Building the .sln for ARM64 inside Visual Studio fails (appears to be the same reasons as win2d.proj)

Build failures in the test and sample projects are from a few different issues:

- Unresolved externals in ExampleGallery.Direct3DInterop.uap, because the directxtk_uwp.2015.7.29.1 package does not support ARM64.  It looks like this is already fixed, so simply upgrading to the latest version of directxtk should solve that.
- CppUnitTestCommon.h rejects ARM64 with #error "Not supported architecture".  Follow up with VS unit test owners about this?  Just disabling all the Win2D tests on ARM64 doesn't seem ideal :-(
- csc produces an error when building the ExampleGallery BackgroundTask and Effects projects: "Invalid option 'ARM64' for /platform; must be anycpu, x86, Itanium or x64".  I'm not sure why the other C# projects do not fail in this same way.
- All the C# projects (samples and tests) fail with: "Your project.json doesn't have a runtimes section. You should add '"runtimes": { "win10-arm64": { } }' to your project.json and then re-run NuGet restore."

jkunkee:
Hi Shawn,

Thanks for giving it a shot!

* Bump the DX TK - easy
* The unit tests were not enabled for ARM, so I had not enabled them for ARM64--but yes, having the unit tests for ARM64 would be best.
* This csc error suggests it's an old version. I'll see if I can repro it and figure out a workaround.
* The project.json and other Nuget errors mystified me. The only solution that I can find online is to transplant the source files into a brand new project, and you know these projects better than I do.

shawnhar:
Afraid I know next to nothing about how project.json works.  None of these projects are even using such a file!   My guess is the VS build system is somehow internally generating a project.json from the information in the .csproj, but getting this wrong due to some detail of how the .csproj is configured.

The (not particularly good) best way I know to debug such things is to create a new working .csproj from a VS project template, then compare its contents with the one I'm trying to upgrade to figure out what the key difference is.

marswe:
To get past the project.json error about win10-arm64, you need to upgrade the PackageReference for Microsoft.NETCore.UniversalWindowsPlatform to version 6.2.2.  Once you do that and build as x86 (for example) to pull down the NuGet package, you'll be able to build ARM64.

jkunkee:
The 6.2.8 bump doesn't change anything here, actually. The Nuget package still builds fine, but the samples still break with the same error:

```
"C:\Users\jkunkee\source\repos\Win2D\Win2D.proj" (default target) (1) ->
"C:\Users\jkunkee\source\repos\Win2D\winrt\test.managed\winrt.test.managed.uap.csproj" (default target) (5:4) ->
(ResolveNuGetPackageAssets target) ->
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(186,5): error : Your project file d
oesn't list 'win10-arm64' as a "RuntimeIdentifier". You should add 'win10-arm64' to the "RuntimeIdentifiers" property in your project file and then re-r
un NuGet restore. [C:\Users\jkunkee\source\repos\Win2D\winrt\test.managed\winrt.test.managed.uap.csproj]

"C:\Users\jkunkee\source\repos\Win2D\Win2D.proj"` (default target) (1) ->
"C:\Users\jkunkee\source\repos\Win2D\winrt\test.managed\winrt.test.managed.uap.csproj" (default target) (5:8) ->
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(186,5): error : Your project file d
oesn't list 'win10-arm64' as a "RuntimeIdentifier". You should add 'win10-arm64' to the "RuntimeIdentifiers" property in your project file and then re-r
un NuGet restore. [C:\Users\jkunkee\source\repos\Win2D\winrt\test.managed\winrt.test.managed.uap.csproj]
```

Incidentally, I ran into this a lot with my other branch: making seemingly necessary changes, like TargetPlatformMinVersion, sometimes didn't seem to fix anything. (I'd like to fix what's in front of me so I don't get bogged down in combinatorics; for example, I tried bumping the MSTest packages too and that led to PCH errors.)

shawnhar:
Are you doing a clean rebuild after this bump?  The C# msbuild system caches a bunch of generated project files in the obj folders, and I've had problems in the past with them not always rebuilding when you edit projects directly rather than via VS UI.

Win2D isolates all outputs into the bin and obj folders, so you can just nuke those two to force a clean build.

jkunkee:
My build sequence in the root of the repo is:

```
git clean -xdf .
build\nuget\download-nuget.cmd
build.cmd OR msbuild /m Win2D.proj /binaryLogger:foo.binlog
```

It's a bit heavy, but it leaves me wondering about fewer variables.

shawnhar:
That'll work for sure.  When testing Win2D.proj, you can get better iteration time by scoping down to just one flavor, passing eg. /p:BuildPlatforms=ARM64 /p:BuildConfigurations=Debug

jkunkee:
I'll give that a try.

I think I'm down to these problems with the tests:

1. It does not seem that MSBuild fully understands conditionally split `TargetPlatformMinVersion` values. Nuget restore produces a project.assets.json file with 16299 while the VS task for asset restoration uses the split value.
2. Win2D's tests appear to use reflection (ReflectionTests?) which is not supported by .NET Native. (This isn't quite right because this would be broken on x64 too, but it's definitely broken on ARM64.)
3. Only ARM64 reports a missing function definition, `System.Void Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProperty.set_Attributes(Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPropertyAttributes)`. This may be due to 2.

jkunkee:
Item 3 was due to item 2: winrt.test.managed.uap also fails if I force x64 to use the .NET Native toolchain. I'm fixing that by simply turning off that project for ARM64, which is where item 1 was coming from.

shawnhar:
Urgh.  We might need Visual Studio help to sort out the conditionally split min version issue?

I think reflection is working in test.managed for other architectures for the simple reason that this project does not enable .NET Native :-)    Is that required for ARM64?   Even if it is, hopefully reflection should still be usable via a suitable .rd.xml configuration.

jkunkee:
Do you know how to author an rd.xml for this situation?

jkunkee:
I'm almost done with my .rd.xml file; I think it's quite a bit cleaner than 8eef14d.

Also, the split version seems to be working, but I don't think I've tried it thoroughly enough. I think ultimately I need to try using the generated Nuget package in a 10240-minversioned x86 project.

jkunkee:
After all that, it looks like MSTest needs to add a .rd.xml file to one of their assemblies. That can't be fixed here.

Until then, would it work to merge what I've already pushed? Bumping MSTest can start with reverting 8eef14d.

jkunkee:
With this last rebase and push, Win2D.proj now contains the complete build graph. Build.cmd and msbuild invocations should now work and Visual Studio will work so long as you rerun Nuget restore when switching to or from ARM64.

shawnhar:
What Visual Studio version are you using for this?   I'm on 2019 v 16.1.1, but can't get the latest to compile. 

build.cmd produces a few instances of this warning:

    Warning	MSB8021	The value 'NotSet' of the variable 'CharacterSet' is incompatible with the value 'true' of the variable 'WindowsAppContainer'.	winrt.lib.uap	C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets	379

before failing with:

    C:\Program Files (x86)\Windows Kits\10\Include\10.0.17134.0\um\windows.h(167): fatal error C1083: Cannot open include file: 'excpt.h': No such file or directory [F:\Win2D\Enlistment\winrt\lib\winrt.lib.uap.vcxproj]

Building ARM64 from win2d.uap.sln, after cleaning the enlistment, produces a set of errors:

    Severity	Code	Description	Project	File	Line
    Warning		The value 'NotSet' of the variable 'CharacterSet' is incompatible with the value 'true' of the variable 'WindowsAppContainer'.	ExampleGallery.Effects		
    Error		Your project file doesn't list 'win10' as a "RuntimeIdentifier". You should add 'win10' to the "RuntimeIdentifiers" property in your project file and then re-run NuGet restore.	winrt.test.managed.uap		
    Warning		The value 'NotSet' of the variable 'CharacterSet' is incompatible with the value 'true' of the variable 'WindowsAppContainer'.	ExampleGallery.Effects		
    Warning		The value 'NotSet' of the variable 'CharacterSet' is incompatible with the value 'true' of the variable 'WindowsAppContainer'.	ExampleGallery.uap		
    Warning		Your project file doesn't list 'win10' as a "RuntimeIdentifier". You should add 'win10' to the "RuntimeIdentifiers" property in your project file and then re-run NuGet restore.	winrt.test.managed.uap		
    Warning		Your project file doesn't list 'win10' as a "RuntimeIdentifier". You should add 'win10' to the "RuntimeIdentifiers" property in your project file and then re-run NuGet restore.	winrt.test.managed.uap		
    Warning	MSB8021	The value 'NotSet' of the variable 'CharacterSet' is incompatible with the value 'true' of the variable 'WindowsAppContainer'.	winrt.test.external.uap	C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets	379
    Error	C1189	#error:  "Not supported architecture"	winrt.test.external.uap	c:\program files (x86)\microsoft visual studio\2017\enterprise\vc\auxiliary\vs\unittest\include\uwp\cppunittestcommon.h	69

Building the .sln for x64, switching to ARM64, running a nuget restore, and then building, produces about 500 errors about being unable to find any of the System.* types.

Any ideas?   I'm hoping I'm just using the wrong VS version here :-)

Couple of minor requests:

* Assuming that VS2019 is now required to build Win2D, update the 2017 references in README.md
* Ditto in build.cmd, plus change the 15.0 check to 16.0

Thanks!

jkunkee:
@shawnhar, I was using VS 2017 15.9.11. I have not tried with VS 2019.

Should I update all the csproj references to 12.0 too?

jkunkee:
The 'unsupported arch' and 'RuntimeIdentifier' issues should be cleared up with a NuGet cache purge. They are due to references to pre-ARM64-support SDK versions.

shawnhar:
I cleaned the enlistment before trying this, which wiped the packages directory.  Do you mean something more by a NuGet cache purge?

I tried with both VS 2017 and 2019.

jkunkee:
Yes, I had to do something along the lines of `rmdir /s /q %HOMEPATH%\.nuget`.

shawnhar:
Even after wiping my system .nuget cache, build.cmd from a VS 2019 developer prompt fails immediately after completing package restore.   I'm stumped on what the difference between our environments might be here?

    Saved output metadata file Microsoft.Graphics.Canvas.winmd.
    Validating metadata file F:\Win2D\Enlistment\build\..\bin\UAPARM64\Debug\MergedWinMD\Microsoft.Graphics.Canvas.winmd.
    pch.cpp
    :\Program Files (x86)\Windows Kits\10\Include\10.0.17134.0\um\windows.h(167): fatal error C1083: Cannot open include file: 'excpt.h': No such file or directory [F:\Win2D\Enlistment\winrt\lib\winrt.lib.uap.vcxproj]
    pch.cpp
    C:\Program Files (x86)\Windows Kits\10\Include\10.0.17134.0\um\windows.h(167): fatal error C1083: Cannot open include file: 'excpt.h': No such file or directory     
    [F:\Win2D\Enlistment\winrt\lib\winrt.lib.uap.vcxproj]
    Build failed; aborting.

jkunkee:
Well, at least it's a different error I suppose. That means the cache purge did help with the other errors.

When I run `msbuild /m Win2D.proj /p:RunTests=false` it succeeds. (I haven't set this machine up to run tests.)

I just searched through my build output and I'm using the same 10.0.17134.0 SDK.

Oh, again, our VS versions are different. I can try with 2019; as I said I have been using 2017.

That said, the error about excpt.h is weird; the Windows SDK should contain that header:

    > dir /s /b "c:\Program Files (x86)\Windows Kits"\excpt.h
    c:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\km\crt\excpt.h

jkunkee:
When I build with the Developer Command Prompt for VS 2019, evaluating the line

    <VsTest>"$([System.IO.Path]::GetFullPath('$(_VsTest)'))"</VsTest>

produces the error

    Win2D\Win2D.proj(273,5): error MSB4184: The expression "[System.IO.Path]::GetFullPath('')" cannot be evaluated. The path is not of a legal form.

Did you have to work past this? I'm working around this by setting RunTests=false the string to an incorrect but 'legal' value.

jkunkee:
When I installed VS2019, it included Windows 10 SDK 10.0.17763.0. I went ahead and checked the box for 10.0.17134.0 thinking to make sure things matched up.

I then repro'd the break you describe, also described [here](https://stackoverflow.com/questions/13656856/cant-get-visual-studio-c-include-file-excpt-h-to-get-installed). It went away when I unchecked SDK 10.0.17134.0 in VS2019--even though the SDK was still checked for VS2017.

shawnhar:
Sorry it's taking me so long to dig into this.   After trying a number of different things, I'm still failing to find a combination of Visual Studio and SDK versions that doesn't suffer from mysterious build errors.   I can get the errors to change around a lot depending on what I have installed, though!   Fear it may be time for a complete uninstall/reinstall.

Can I confirm what version you are currently using, and that both "msbuild win2d.proj" and Win2D.uap.sln builds inside the VS IDE are succeeding?

azchohfi:
The fix for the GetFullPath not working is to just add an additional line for VS**160**COMNTOOLS:
`	<_VsTest Condition="Exists('$(VS160COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS160COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
`
After doing that, I'm hitting the excpt.h problem as well.
Updating to toolset 142 fixes it and it generates the package just fine.

azchohfi:
I had to cherry-pick a couple of commits from #711. Should we merge that one before we merge this?

shawnhar:
I think that's a good idea.  The remaining blocker on that PR was winrt.test.managed.uap not being able to run tests any more, but it looks like you figured out why this was?

lhak:
Maybe the xaml_island branch can also be merged now (it also contains support for Visual Studio 2019)?

shawnhar:
This is now merged into master via https://github.com/microsoft/Win2D/pull/722, so this PR is no longer needed.

Issue #700, opened 2019-05-14
CanvasControl does not trigger Draw if it was wrongly onloaded by changing it's parent


victorpadureanu:
CanvasControl relied on the Loaded and Unloaded Event. It will call the draw function is the canvas was loaded. If it sees that canvas as being unloaded ( even if it is visible on screen and I call Invalidate ) it will not call draw.
I managed to repro this behavior with a simple button:
[https://github.com/victorpadureanu/LoadedUnloadedIssue](https://github.com/victorpadureanu/LoadedUnloadedIssue)

Do you know a reliable way of changing the parent of the CanvasControl / forcing it to call draw?

rosca:
Related discussion on Microsoft forums: https://social.msdn.microsoft.com/Forums/windowsapps/en-US/ef9c34c4-8bd0-4a43-b9ec-46ee1a70a06f/why-are-there-more-unloaded-than-loaded-events?forum=wpdevelop#ef9c34c4-8bd0-4a43-b9ec-46ee1a70a06f

shawnhar:
I can't think of any way to make Win2D completely robust against the current behavior of these events, so believe it will require a XAML fix in the OS to solve this.

The problem is that XAML can not only sometimes deliver unloaded event without a matching loaded, but can also deliver loaded and unloaded out of order.  So when Win2D receives an unloaded event followed by loaded, it's impossible to tell whether this is due to

1) A control being added, then removed before the loaded event had fired, then added again (in which case it should be visible)
2) A control being added and immediately removed, with the two events delivered out of order (in which case it should be invisible)

The Win2D control implementation maintains a loaded/unloaded counter that makes it robust against the later case.  Changing this behavior to fix the former would break the latter, which doesn't seem like the right tradeoff.

PR #704, opened 2019-06-06
 Use VisualTreeHelper.GetParent instead of a counter to determine Loaded/Unloaded

victorpadureanu:
Fix for #700

### DIFF
--- a/winrt/lib/xaml/BaseControl.h
+++ b/winrt/lib/xaml/BaseControl.h
@@ -132,6 +132,13 @@ namespace ABI { namespace Microsoft { namespace Graphics { namespace Canvas { na
 
         ComPtr<Media::IVisualTreeHelperStatics> m_VisualTreeHelper;
 
+        enum class LoadAction
+        {
+            kNoOp = 0,
+            kLoad = 1,
+            kUnload = 2
+        };
+
     public:
         BaseControl(std::shared_ptr<adapter_t> adapter, bool useSharedDevice)
             : m_adapter(adapter)
@@ -792,17 +799,7 @@ namespace ABI { namespace Microsoft { namespace Graphics { namespace Canvas { na
             return ExceptionBoundary(
                 [&]
                 {
-                    boolean should_load = false;
-                    if (WasLoaded())
-                    {
-                        should_load = true;
-                        auto lock = GetLock();
-                        if (m_isLoaded) {
-                            should_load = false;
-                        }
-                    }
-
-                    if (should_load)
+                    if (GetLoadAction() == LoadAction::kLoad)
                     {
                         RegisterEventHandlers();
                         UpdateIsVisible();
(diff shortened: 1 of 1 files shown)


shawnhar:
With this change, 218 of the 1686 tests in winrt.test.internal are failing.

I expect a couple of the tests will need updating to match the new behavior, but also it looks like the unit test mocks need changing to support this newly added IVisualTreeHelperStatics usage.

winrt/lib/xaml/BaseControl.h
```diff
@@ -129,11 +130,17 @@ namespace ABI { namespace Microsoft { namespace Graphics { namespace Canvas { na
 
         ComPtr<ICanvasDevice> m_customDevice;
 
+        enum class LoadAction
+        {
+            kNoOp = 0,
```
shawnhar: nit: remove the k prefix, to match conventions of other Win2D code.

victorpadureanu:
I used both behaviors and if any of them says loaded it will be loaded.

shawnhar:
Using both behaviors side by side isn't ideal, because it means the tests will not be validating the same code paths that are actually used when Win2D is running inside an app.  This makes the tests significantly less valuable.

The approach used by Win2D is to indirect via adapter interfaces whenever Win2D needs to interact with its host environment.  This lets the unit tests provide emulated versions of whatever XAML functionality Win2D depends on, and mock the behavior of the XAML framework so as to exercise all the possible code paths through Win2D itself in the same way that a real XAML environment would.

For BaseControl, this is done using [IBaseControlAdapter](https://github.com/microsoft/Win2D/blob/master/winrt/lib/xaml/BaseControl.h#L28).  That provides abstract definitions of all the places BaseControl needs to interact with things like its window, UI thread dispatcher, DPI changed event notification, etc.  It can be extended to include access to the visual tree helper (or perhaps just the single GetParent method that this change requires).   BaseControl then uses m_adapter for all access to these features of its outside environment.

In normal usage, IBaseControlAdapter is implemented by [BaseControlAdapter](https://github.com/microsoft/Win2D/blob/master/winrt/lib/xaml/BaseControlAdapter.h#L17), which does a trival pass-through of everything to their real XAML implementations (see GetWindowOfCurrentThread for a simple example).

For unit tests, IBaseControlAdapter is implemented by [BaseControlTestAdapter](https://github.com/microsoft/Win2D/blob/master/winrt/test.internal/xaml/BaseControlTestAdapter.h#L14) and its various subclasses, which allow unit tests to artificially inject whatever result they want the control to observe (see for example the BaseControlTestAdapter implementation of SetHasUIThreadAccess).

Hope this explanation helps, and my apologies for not having this architectural information documented properly anywhere :-(

shawnhar:
A different solution for this issue was incorporated into Win2D release 1.25.0, so this PR is no longer needed

PR #711, opened 2019-07-19
Fix build with Visual Studio 2019

wjk:
Does what the title says, mostly. I did not touch any of the test projects, because I didn't want to accidentally break them.

### DIFF
--- a/Win2D.proj
+++ b/Win2D.proj
@@ -233,7 +233,7 @@
     <_VsTest Condition="Exists('$(VS120COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS120COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
     <_VsTest Condition="Exists('$(VS140COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS140COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
     <_VsTest Condition="Exists('$(VS150COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS150COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
-    <VsTest>"$([System.IO.Path]::GetFullPath('$(_VsTest)'))"</VsTest>
+    <VsTest Condition="'$(_VsTest)' != ''">"$([System.IO.Path]::GetFullPath('$(_VsTest)'))"</VsTest>
   </PropertyGroup>
 
   <Target Name="RunTests"
@@ -242,6 +242,7 @@
           Inputs="@(TestProjects)"
           Outputs="%(TestProjects.TestAppX)">
 
+    <Error Text="vstest.console.exe not found, please specify using the VsTest property" Condition="'$(VsTest)' == ''" />
     <Message Importance="High" Text="Running tests: %(TestProjects.Filename) (%(TestProjects.Configuration)|%(TestProjects.Platform))" />
 
     <Exec Command="$(VsTest) %(TestProjects.TestBinary)%(TestProjects.TestArgs)" ContinueOnError="ErrorAndContinue" IgnoreStandardErrorWarningFormat="true" />
--- a/README.md
+++ b/README.md
@@ -69,14 +69,14 @@ Studio, add the Win2D NuGet package to your project, and get started using the A
 ## Building Win2D from source
 
 ##### Requirements
-- Visual Studio 2017 with Tools for Universal Windows Apps 15.0.27428.01 and Windows SDK 17134
+- Visual Studio 2019 with Tools for Universal Windows Apps 15.0.27428.01 and Windows SDK 18362
 
 ##### Clone Repository
 - Go to 'View' -> 'Team Explorer' -> 'Local Git Repositories' -> 'Clone'
 - Add the Win2D repository URL (https://github.com/Microsoft/Win2D.git) and hit 'Clone'
 
 ##### Build NuGet Packages
-- Launch 'Developer Command Prompt for VS2017'
+- Launch 'Developer Command Prompt for VS2019'
 - Change directory to your cloned Win2D repository and run 'build'
 
 ##### Point Visual Studio at the resulting 'bin' directory
--- a/Win2D.uap.sln
+++ b/Win2D.uap.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2010
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29020.237
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "winrt", "winrt", "{0F212400-0A8B-4E26-9BA8-CE7165FF8717}"
 EndProject
--- a/build.cmd
+++ b/build.cmd
@@ -2,7 +2,7 @@
 
 SETLOCAL
 
-IF "%VisualStudioVersion%" LSS "15.0" (
+IF "%VisualStudioVersion%" LSS "16.0" (
     GOTO WRONG_COMMAND_PROMPT
 )
 
@@ -35,7 +35,7 @@ GOTO END
 
 :WRONG_COMMAND_PROMPT
 
-ECHO Please run this script from a Developer Command Prompt for VS2017
+ECHO Please run this script from a Developer Command Prompt for VS2019
 ECHO.
 PAUSE
 GOTO END
--- a/build/Win2D.cpp.props
+++ b/build/Win2D.cpp.props
@@ -45,7 +45,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>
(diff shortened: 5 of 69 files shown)


shawnhar:
Hi,  I get a bunch of errors from mdmerge trying to build this:

MDMERGE : error MDM2012: Error 3 has occured while enumerating files in C:\Program Files (x86)\Windows Kits\10\References\10.0.17763.0\Windows.Foundation.UniversalApiContract\8.0.0.0\*.winmd. [F:\Win2D\Enlistment\winrt\lib\winrt.lib.uap.vcxproj]

Looks like this is because my folder C:\Program Files (x86)\Windows Kits\10\References\10.0.17763.0\Windows.Foundation.UniversalApiContract contains 7.0.0.0 rather than 8.0.0.0.

Any idea what that's about?

wjk:
I saw that before, and fixed it in efa33a5. I retargeted the project to the 18362 SDK, which includes UniversalApiContract 8.0.0.0. If you try building using that version of the Windows SDK, it should work. If you know of a good way to make this work for any SDK version, not just this specific one, please let me know. Thanks!

shawnhar:
Hmm, something is odd here!  Your commit changed the number from 6.0.0.0 to 8.0.0.0, but on my machine what I actually have installed is 7.0.0.0.

I think this is because your change sets WindowsTargetPlatformVersion to 10.0 in Win2D.cpp.props, rather than specifying the specific version that Win2D is set up to build with.  That means to use whatever the latest version that happens to be installed is, which might fail due to either being older or newer than what the Win2D codebase expects.

Since the Win2D build is not resilient against changing SDK versions (for reasons like this path information), it should be explicit about what SDK is to be used.

shawnhar:
Another issue with this change is that winrt.test.internal.uap.vcxproj is still using the older PlatformToolset = v141.  This should be kept in sync with the definition from Win2D.cpp.props.

shawnhar:
Another needed change:  Win2D.uap.sln should update the Visual Studio version information at the top of the file, so it will open in VS 2019 rather than 2017 if both are installed.

Thanks!

wjk:
@shawnhar All requested changes have been implemented; however, when I tried changing the UniversalApiContract version to 7.0.0.0 per your specification, it fails to build on my computer. The only version of that contract I have is 8.0.0.0. I don't know what could be going on here.

shawnhar:
Thanks for the updates!  I'm testing this out now.

My error about the UniversalApiContract was because I didn't actually have the right Windows SDK installed on my dev box, but before your change to make the SDK version explicit, it would just build using whatever the latest I had installed was instead, rather than giving a helpful error message for that installation mismatch.

wjk:
@shawnhar One other question for you. About how long after this PR gets merged will it take for a new release containing these changes and those from #712 to show up on nuget.org? The need to use a private build of this library is the main thing stopping me from publishing some code I was working on to bring the UWP Acrylic effect to a Windows Forms app, no XAML Island required. Thanks!

shawnhar:
With the right SDK installed I can successfully run build.cmd, but am getting errors when trying to build the full test and sample code via "msbuild /m win2d.proj" or loading Win2D.uap.sln and doing an F5 to build the whole thing.   Failures:

1) winrt.test.nativecomponent his compile errors in ShaderCompiler.cpp.  This was already fixed in the [xaml_islands](https://github.com/microsoft/Win2D/blob/xaml_islands/winrt/test.nativecomponent/ShaderCompiler.cpp) branch - that fix will need pulling across here for compatibility with the latest Visual Studio.

2) I get a warning about needing to retarget the platform toolset when loading winrt.test.internal.uap.vcxproj into the IDE, and then a build failure when compiling it.  Not sure why, but apparently the platform toolset needs to be set explicitly in this project rather than just inheriting it from the common props.

3) The C# sample projects fail to build with errors of the form:

    SimpleSample
    1>------ Build started: Project: SimpleSample.uap, Configuration: Debug Win32 ------
    1>F:\Win2D\Enlistment\samples\SimpleSample\SimpleSample.uap.csproj : XamlCompiler error WMC1006: Cannot resolve Assembly or Windows Metadata file 'System.Runtime.dll'
    1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\WindowsXaml\v16.0\8.2\Microsoft.Windows.UI.Xaml.Common.targets(266,5): Xaml Internal Error error WMC9999: Type universe cannot resolve assembly: System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a.
    ========== Build: 0 succeeded, 1 failed, 2 up-to-date, 0 skipped ==========

Couple of other issues I noticed:

* The unit tests in tools\tools.sln will fail, because this PR manually changes the generated effects code without updating the tool that generated it.   tools\codegen\exe\OutputEffectType.cs needs a matching update for this.

* The VS 2017 references in README.md are out of date

* Lines 5 and 38 of build.cmd are out of date

Thanks for working on this, and sorry for the barrage of feedback!

shawnhar:
Is everything building successfully for you now?  With the latest changes I am still hitting some (different) compile errors.

wjk:
@shawnhar The build now succeeds with no errors on my computer. I'll keep you abreast of any test errors that occur.

wjk:
@shawnhar The unpackages tests all pass; I cannot run the AppX tests due to timeout failures connecting to the packaged process (it deploys and launches the app, it just sits there with a blank screen, and after 90 seconds it quits the app and vstest.console.exe emits an error). If you could test these on your machine, I would be appreciative. Thanks!

shawnhar:
This is getting close!   Everything builds for me now, and the winrt.test.internal.uap and winrt.test.external.uap tests are passing both inside the IDE and when building win2d.proj.   I see the same thing as you where winrt.test.managed.uap is failing, though not quite the same symptom.  Looking at the Tests output pane in VS, it seems the error is due to a misconfiguration of this project:

    [7/25/2019 10:32:49 AM Error] The project F:\Win2D\Enlistment\winrt\test.managed\winrt.test.managed.uap.csproj does not have a reference to Microsoft.Net.Test.Sdk nuget package. For more information visit https://go.microsoft.com/fwlink/?linkid=2028865.
    [7/25/2019 10:32:51 AM Informational] ------ Discover test started ------
    [7/25/2019 10:33:04 AM Informational] ========== Discover test finished: 1686 found (0:00:13.0511669) ==========

I'm also seeing a strange thing where if I clean the enlistment, then load the .sln into the VS UI, the first time I build the managed projects all fail with:

    Error	CS0246	The type or namespace name 'System' could not be found (are you missing a using directive or an assembly reference?)	CompositionExample, CoreWindowExample.uap, ExampleGallery.BackgroundTask.uap, ExampleGallery.Effects	F:\temp\.NETCore,Version=v5.0.AssemblyAttributes.cs	2	N/A

Closing and reopening the .sln fixes this.   I have no idea what is going wrong here but suspect a Visual Studio bug rather than Win2D issue, so it's probably ok to ignore this.

Win2D.proj
```diff
@@ -233,7 +233,7 @@
     <_VsTest Condition="Exists('$(VS120COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS120COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
     <_VsTest Condition="Exists('$(VS140COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS140COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
     <_VsTest Condition="Exists('$(VS150COMNTOOLS)$(_VsTestPathUnderCommonTools)')">$(VS150COMNTOOLS)$(_VsTestPathUnderCommonTools)</_VsTest>
```
azchohfi: This should be adding a new condition for VS**160**COMNTOOLS

shawnhar: Nice catch!  I will test this out tomorrow.

azchohfi: It executed all tests for me

azchohfi: It actually required a few more tweaks since everything was updated. I'll push a new branch out of this one with a single new commit.

azchohfi: Now it's running the tests the first time, but then it tries to build for different arch (x64 x Win32) and fails to run the tests because the appx with the same id already exists... It's not uninstalling or replacing the version with the new one, so I reduced from 4 bugs to 3. I've found this: https://github.com/Microsoft/vstest/issues/1520 Investigating...

shawnhar: I see the same thing trying this locally.   The workaround mentioned in that vstest issue of manually uninstalling the package before attempting to run tests seems plausible?

azchohfi: Not really, because the whole process is automated. This will never run on a vanilla build server without changing the build script. We need to either automatically try to uninstall the package before running the tests for each arch/config or change the generated package name for each arch/config. What do you think is the best solution?

shawnhar: Yep, this would need to be automated in Win2D.proj to attempt an uninstall before running each test pass.

I think changing generated package names would still run into the same bug if more than one build was run back-to-back on the same machine?

azchohfi: No, they would replace the old installation. The problem is that the IDs are different but the names are the same. It's not hard to remove it from PS:
get-appxpackage -name E2C40EB9-8CCE-48B1-9A5C-4C8CA4999631 | remove-appxpackage
I'll send the fix.

azchohfi: I sent #721 that has all the commits here and the tests works.

shawnhar:
These changes have now been merged via PR https://github.com/microsoft/Win2D/pull/721, where azchohfi added an additional commit to get the managed unit tests working again.   I'm closing this PR because its contents were already merged via that alternate route.

PR #712, opened 2019-07-19
Add support to netcoreapp3.0 projects to NuGet package

wjk:
This will aid consuming Win2D in projects [such as this one](https://github.com/microsoft/Windows.UI.Composition-Win32-Samples/tree/master/dotnet/WinForms/AcrylicEffect), which currently requires manually unpacking the Win2D nuget package and coping a file into the bin directory, which is hard to automate using a build system (at least without checking prebuilt binary files into your repo). I can also add .NET Framework support if desired; if so, please let me know what TFM I should use. Thanks!


shawnhar:
This looks good to me; thanks for putting it together!

@damyanp would appreciate your eyes on this as well please?

PR #718, opened 2019-08-21
Corrected the method name.

goh-chunlin:
The method name in XAML is canvas_DrawAnimated but backend code is canvas_AnimatedDraw. This PR is to correct the backend code.


shawnhar:
Good catch, thank you for correcting this!

PR #723, opened 2019-09-10
[Draft] Xaml islands

azchohfi:
This works but it's not the right solution.
I spoke with @marb2000 and the right solution is to provide some other mechanism for apps to detect when the Islands are visible or not.


shawnhar:
After installing the .NET Core SDK I can build this, but am getting a bunch of warnings, the first of which fails the build due to treat-warnings-as-errors:

    F:\Win2D\Enlistment\winrt\test.internal\mocks\MockD2DFactory.h(287,17): warning C4193:  #pragma warning(pop): no
        matching '#pragma warning(push)' [F:\Win2D\Enlistment\winrt\test.internal\winrt.test.internal.uap.vcxproj]

    F:\Win2D\Enlistment\samples\ExampleGallery\Effects\ExampleGallery.Effects.csproj(20,3): warning MSB4011: "F:\W
       in2D\Enlistment\build\Win2D.cs.props" cannot be imported again. It was already imported at "F:\Win2D\Enlistment\
       Directory.Build.props (56,3)". This is most likely a build authoring error. This subsequent import will be ignor
       ed. [F:\Win2D\Enlistment\Win2D.proj]

shawnhar:
This work was incorporated into Win2D release 1.25.0, so this PR is no longer needed

PR #767, opened 2020-05-08
Fixed typo

r2d2rigo:


shawnhar:
Good catch, thanks for the fix!

Issue #799, opened 2020-11-30
Suggestion: m_lastSeenParent should be weak reference (would you accept a PR?)


benstevens48:
Following on from this issue <https://github.com/microsoft/Win2D/issues/658> and by looking at the code, I can see that Win2D controls hold a strong reference to their parent (called m_lastSeenParent). This is to enable RemoveFromVisualTree to work correctly. 

I propose changing m_lastSeenParent to a WeakRef. Would you accept a PR for this?

**Reasoning**
The current behavior causes a number of problems. As mentioned in the other issue, even unsubscribing all events from the control will not prevent a memory leak because there is this strong reference to the parent remaining. Also, this means that RemoveFromVisualTree must even be called in C++ apps. Both of these are in contradiction to the docs [here](https://microsoft.github.io/Win2D/html/RefCycles.htm).

Also, it's not always feasible to have a `Dispose`-type method on the containing control - in my case I am using CanvasVirtualControl in a reusable component,  and requiring a Dispose on this would require a Dispose on all possible parent controls. What I want to do is unsubscribe events on Unloaded and subscribe them on Loaded, but this won't prevent a memory leak due to the strong parent ref held by the CanvasVirtualControl. (Also, holding a strong parent ref after unloading from the visual tree is the exact opposite of what the framework controls do. It makes more sense to me that the child should hold no strong refs to the parent, rather than trying to do it the other way around).

benstevens48:
I created a pull request #801 for this.

PR #801, opened 2020-12-04
Change last seen parent to weak ref

benstevens48:
As discussed in issue <https://github.com/microsoft/Win2D/issues/799>.

One question - if the parent is null in UpdateLastSeenParent, should we keep the old parent weak reference? At the moment I'm resetting it since that is the same as the current behavior. ie. should I remove the line `m_lastSeenParent.Reset();`?


JoseIvanHernandez (approved):
Commit

PR #807, opened 2021-02-17
Add thread safety to WIC initialization using call_once

benstevens48:
Technically the singleton DefaultWicAdapter instance was not thread-safe when creating the WIC factory. Similarly, the DefaultCanvasImageAdapter was not thread-safe when storing a copy of the WicAdapter instance. In the second case in practice it probably was thread-safe anyway assuming setting a pointer is atomic. In the first case it would probably have been very difficult to hit a serious issue, although potentially it could have leaked the factory object and potentially returned a different instance on different calls.

This pull request fixes that by using std::call_once.


damyanp (approved):

MartyIX:
Will this get merged please?

PR #902, opened 2023-02-01
Upgrade toolchain to v143

Dev-Taehui:
Use Visual C++ toolchain v143 instead of v142.


Dev-Taehui:
@microsoft-github-policy-service agree

Sergio0694:
We've already bumped to v143 internally, this will be included in the 1.27.0-preview2 changes 🙂

PR #903, opened 2023-02-01
Set Direct3D feature level 11.1

Dev-Taehui:
Without Direct3D feature level array, I found that out variable (pFeatureLevel) is D3D_FEATURE_LEVEL_11_0.

In the official Direct2D documentation, featureLevels were created from 11.1 to 9.1 and provided to D3D11CreateDevice. This results out variable pFeatureLevel sets D3D_FEATURE_LEVEL_11_1.

I don't know how this will work, but it doesn't look bad using 11.1.


Dev-Taehui:
I rebased to winappsdk/main.

PR #904, opened 2023-02-01
Support VRR (Variable Refresh Rate)

Dev-Taehui:
To support VRR, these will be set.

1. Set SwapChain allow tearing
2. Provide DXGI_PRESENT_ALLOW_TEARING to Present function


Dev-Taehui:
Close for renaming branch name

PR #906, opened 2023-02-01
Expose DXGI device for some usage (NVIDIA(R) Reflex(TM), etc.)

Dev-Taehui:
Exposing DXGI device enables to implement some technologies such as NVIDIA(R) Reflex(TM).

### DIFF
--- a/winrt/lib/drawing/CanvasDevice.abi.idl
+++ b/winrt/lib/drawing/CanvasDevice.abi.idl
@@ -103,7 +103,7 @@ namespace Microsoft.Graphics.Canvas
         [propget] HRESULT LowPriority([out, retval] boolean* value);
         [propput] HRESULT LowPriority([in] boolean value);
 
-        [propget] HRESULT DXGIDevice([out, retval] IUnknown** dxgiDevice);
+        [propget] HRESULT DXGIDevice([out, retval] IInspectable** dxgiDevice);
 
         //
         // This event is raised whenever the native device resource is lost-
--- a/winrt/lib/drawing/CanvasDevice.abi.idl
+++ b/winrt/lib/drawing/CanvasDevice.abi.idl
@@ -103,6 +103,8 @@ namespace Microsoft.Graphics.Canvas
         [propget] HRESULT LowPriority([out, retval] boolean* value);
         [propput] HRESULT LowPriority([in] boolean value);
 
+        [propget] HRESULT DXGIDevice([out, retval] IUnknown** dxgiDevice);
+
         //
         // This event is raised whenever the native device resource is lost-
         // for example, due to a user switch, lock screen, or unexpected
--- a/winrt/lib/drawing/CanvasDevice.abi.idl
+++ b/winrt/lib/drawing/CanvasDevice.abi.idl
@@ -103,7 +103,7 @@ namespace Microsoft.Graphics.Canvas
         [propget] HRESULT LowPriority([out, retval] boolean* value);
         [propput] HRESULT LowPriority([in] boolean value);
 
-        [propget] HRESULT DXGIDevice([out, retval] IInspectable** dxgiDevice);
+        [propget] HRESULT DXGIDevice([out, retval] UINT64* dxgiDevice);
 
         //
         // This event is raised whenever the native device resource is lost-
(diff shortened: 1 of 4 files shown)


Dev-Taehui:
It doesn't work. (I changed IUnknown* to void*)

Sergio0694:
You can't return an `IUnknown*` (or really any pointer type) from a WinRT API, that's by design.
Only `IInspectable` objects are allowed.

Dev-Taehui:
@Sergio0694 Can I cast `ComPtr<IDXGIDevice3>` to `IInspectable`? I tried `m_dxgiDevice.EnsureNotClosed().CopyTo(dxgiDevice);`

Dev-Taehui:
@Sergio0694 I casted to UINT64, And cast it to IntPtr on C# code. And, it works fine! (NVIDIA Reflex works)

Sergio0694:
I'm quite confident this PR would never get merged. Returning an `UINT64` to snuggle a COM pointer is absolutely not a good API design, and it's also inherently dangerous. It also means callers from a managed context can't benefit from automatic marshalling. The WinRT layer is meant to be safe and easy to use, and this completely defeats that.

But most importantly, the feature this PR is trying to add already exists. If you want to get the underlying DXGI device, you can do so from the interop APIs that are available (they're documented in the header that ships with Win2D). Specifically:
- Get a `CanvasDevice` object
- `QueryInterface` for [`IDirect3DDxgiInterfaceAccess`](https://learn.microsoft.com/windows/win32/api/windows.graphics.directx.direct3d11.interop/ns-windows-graphics-directx-direct3d11-interop-idirect3ddxgiinterfaceaccess)
- Call [`IDirect3DDxgiInterfaceAccess::GetInterface`](https://learn.microsoft.com/windows/win32/api/windows.graphics.directx.direct3d11.interop/nf-windows-graphics-directx-direct3d11-interop-idirect3ddxgiinterfaceaccess-getinterface) passing `IDXGDevice3` or some other IID.

There you go, you have your DXGI device, and no new APIs are needed 😄

FYI @getrou I think this one should be closed.

Dev-Taehui:
@Sergio0694 To be honest, I think so. It seems to be very dangerous for referencing raw pointer. But, I am very new on C++ (I am C#, TypeScript and Kotlin developer). So, I want to just know how to get raw pointer.

I totally agree that this PR should be closed. By the way, does your solution work on C# project?

Sergio0694:
> "It seems to be very dangerous for referencing raw pointer."

Note that your current solution is not so much dangerous as it's just not correct. This line is wrong:

https://github.com/microsoft/Win2D/blob/4c490f4dc18c38451b224d614acacfc2282dc9b2/winrt/lib/drawing/CanvasDevice.cpp#L662

You're returning a COM object without incrementing its reference count. This means that if the device is released after you got that, you could crash due to an access violation exception when trying to use the returned object. The correct way is for the API to use `CopyTo`, which will correctly increment the reference count of the object before returning it. If you use `IDirect3DDxgiInterfaceAccess` as I mentioned, that's already handled for you 🙂

> "does your solution work on C# project?"

Sure! The easiest way is for you to define a `[ComImport]` interface. You can use it as follows:

```csharp
private static void UseDxgiDevice(CanvasDevice canvasDevice)
{
    Guid uuidOfIDxgiDevice3 = new("6007896c-3244-4afd-bf18-a6d3beda5023");

    // Get the DXGI device from the COM interface
    ((IDirect3DDxgiInterfaceAccess)(object)canvasDevice).GetInterface(ref uuidOfIDxgiDevice3, out object dxgiDevice3);

    // Get the raw COM pointer for the DXGI device
    IntPtr dxgiDevice3Ptr = Marshal.GetIUnknownForObject(dxgiDevice3);

    try
    {
        // Use the pointer here (eg. pass to nvidia reflex, etc.)
    }
    finally
    {
        // Release the object when we're done with it
        Marshal.Release(dxgiDevice3Ptr);
    }
}

[ComImport]
[Guid("A9B3D012-3DF2-4EE3-B8D1-8695F457D3C1")]
[InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
internal interface IDirect3DDxgiInterfaceAccess
{
    void GetInterface(ref Guid iid, [MarshalAs(UnmanagedType.IUnknown)] out object p);
}
```

This does exactly what you wanted, with no new Win2D APIs needed 😄

Dev-Taehui:
@Sergio0694 Wow, it works! I will close this PR.

PR #980, opened 2025-01-10
Fix local build and tests for UWP

benstevens48:
I was struggling to get a local build and tests working for UWP. Some of this may have been due to my NuGet being out of data, but in any case, relaying on packages with vulnerabilities and private repositories (presumably as a workaround for that) is not great, even if just for the tests, so I made some changes to get the build and tests working without this.

**Warning: this changes the min Windows version to Windows 10 build 16299 (aka 1709).** Not sure if you can accept this change or not.

A summary of changes
1. Change the min Windows 10 build to 16299 and therefore remove the separate build stage for this.
2. Upgrade all NuGet packages to latest supported versions. Note: the latest supported version of Microsoft.NET.Test.Sdk is 17.3.3 (for UWP in general) and the latest supported version of directxtk_uwp (for ARM32) is 2023.4.28.1 (the latter didn't need upgrading really but I thought I may as well).
3. Remove custom NuGet repositories.
4. Remove the NuGet restore condition != AnyCPU since that prevented restore for the projection project.
5. Add additional app assets that are apparently now required to the test.managed project (actually there may have been some project properties that could be set instead but this way also works).
6. Add the <AppxBundle>Never</AppxBundle> property to the test.managed project otherwise it switches to generating a bundle file after being opened in Visual Studio.
7. Change the publisher to CN=Win2DTemporaryKey to avoid a lot of warnings (not essential but nice to avoid the warnings).

Please ignore the individual commits since I went back and forth several times due to issues like the build randomly switching between generating appx and appxbundle.

As a result of the changes, the tests all pass and there are 5 warnings. 3 warnings that say "warning LNK4075: ignoring '/EDITANDCONTINUE' due to '/OPT:ICF' specification" which I don't really want to touch since I don't have a good knowledge of C++ build args and this is from the main project. Then 2 warnings regarding the projection being experimental.


nreh:
Even though this was never merged in, it helped me build and get intellisense working on my machine so thank you!

Issue #998, opened 2025-09-10
Update Win2D to Support WinAppSDK v1.8 Meta Package


ghost1372:
Hi @Sergio0694 ,
WinAppSDK v1.8 has been released and it introduces a new meta package. Could you please upload a new version of Win2D that targets v1.8? At the moment, we’re unable to install Win2D in a library or app that uses the WinAppSDK v1.8 meta packages.

https://github.com/microsoft/WindowsAppSDK/issues/5811

ghost1372:
Version 1.4.0 is available now with meta package feature

PR #1012, opened 2026-02-12
Update to Windows App SDK 1.8

### LINKED ISSUE #1000: Update to WinAppSDK 1.8
Could you please re-release this built again 1.8 and not referencing the metapackage but just referencing Microsoft.WindowsAppSDK.WinUI ? This is needed for any other 3rd party library that depends on this and trying to avoid taking a dependency on all of the winappsdk (like .NET MAUI)

### LINKED ISSUE #11002: Please update the WindowsAppSdk package version in Win2D.
### Describe the bug

In the testing program, I added a separate `Microsoft.WindowsAppSDK.WinUI` package. However, Win2D has not been updated to a separate metadata package version, resulting in my release package being extremely large, including native DLLs related to ML.

### Why is this important?

This is particularly important when using Win2D and Windows UI alone. I don't need to use ML to expand this type of program.

### Steps to reproduce the bug

As mentioned above

### Actual behavior

Error

### Expected behavior

Normal AOT compilation

### Screenshots

Image

### NuGet package version

WinUI 3 - Windows App SDK 1.8.3: 1.8.251106002

### Windows version

Windows 11 (24H2): Build 26100

### Additional context

_No response_

michael-hawker:
- Bump version to 1.4.0
- Uses Microsoft.WindowsAppSDK.WinUI meta package only version 1.8.260204000
- Removes explicit WebView2 dependency
- Updates Microsoft.Windows.SDK.BuildTools to 10.0.28000.1-RTM
- Updates CsWinRT to 2.2.0 and centralizes as build prop in WindowsAppSdk.common.props

I wasn't able to get the build script to run locally, seems like it's pinging an ADO that may not be turned on atm? Figured I could try at least creating the draft PR here and see what happens.

FYI @Sergio0694, think you may have helped with this the last time, so let me know if any of this looks weird or if you have tips on building the repo.

Fixes #1000, #998
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/11002

This also moves from the whole WindowsAppSdk Package to just the WinUI package, which I think is the main requirement for Win2D? It also gets rid of the WebView2 package which is transitively included through WinUI. I assume was included here earlier to workaround prior migration issues in 1.6 but isn't actually required of Win2D.


benstevens48:
This won't work since the C++ projects require all the transitive dependencies to be added. However, you should be able to just update the nuspec file only and get rid of all the other changes. This relies on binary compatibility, which is discussed here -  https://github.com/microsoft/WindowsAppSDK/discussions/6250 - I am pretty sure 1.8 will not have removed and APIs used by Win2D so it should be fine that Win2D is built against 1.6 but uses 1.8 for the runtime (and in fact no-one has thought to worry about this before). Alternatively, I have a full local build fix (including tests and docs) here - https://github.com/microsoft/Win2D/pull/1014.

